PR #5, opened 2015-12-17
Map controller/add climo periods

basilveerman:
Allows the user to select which dataset is displayed from the currently filtered model set (filteredMeta). Initial state is now generated from the filtered model set received.

### DIFF
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -12,6 +12,13 @@ import styles from './AppController.css';
 
 var App = React.createClass({
 
+  /**
+   * Initial state set upon metadata returning in {@link App#componentDidMount}.
+   * Includes:
+   *   - model_id
+   *   - variable_id
+   *   - experiment
+   */
   getInitialState: function() {
     return {
       meta: []
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -14,12 +14,7 @@ var App = React.createClass({
 
   getInitialState: function() {
     return {
-      meta: [],
-      filter: {},
-      model_id: '',
-      variable_id: '',
-      experiment: '',
-      area: undefined
+      meta: []
     };
   },
 
@@ -79,11 +74,11 @@ var App = React.createClass({
   },
 
   render: function() {
-    var filteredMeta = this.getfilteredMeta()
 
     var getThings = function(thing) {
       return _.unique(this.state.meta.map(function(el){return el[thing]}))
     }.bind(this);
+
     return (
       <Grid fluid={true}>
         <Row>
@@ -102,7 +97,7 @@ var App = React.createClass({
             <div className={styles.map}>
               <MapController
                 variable={this.state.variable_id}
-                meta = {filteredMeta}
+                meta = {this.getfilteredMeta()}
                 onSetArea={this.handleSetArea} />
             </div>
           </Col>
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -61,15 +61,18 @@ var App = React.createClass({
     this.setState({area: wkt});
   },
 
+  filteredMeta: function() {
+    var l = this.state.meta.filter( function(x) {
+      return x.model_id === this.state.model_id && x.experiment === this.state.experiment && x.variable_id === this.state.variable_id
+    }, this );
+  },
+
   findUniqueId: function() {
-      var l = this.state.meta.filter(
-	  function(x) {
-	      return x.model_id === this.state.model_id && x.experiment === this.state.experiment && x.variable_id === this.state.variable_id
-	  }, this
-      );
-      if (l.length > 0) {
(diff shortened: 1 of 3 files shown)


src/components/AppController/AppController.js
```diff
@@ -14,12 +14,7 @@ var App = React.createClass({
 
   getInitialState: function() {
     return {
-      meta: [],
-      filter: {},
-      model_id: 'cgcm3',
-      variable_id: 'tasmax',
-      experiment: 'rcp45',
-      area: undefined
+      meta: []
```
jameshiebert: So, I know it's a bit verbose and functionally unnecessary to declare our state variables here as empty or undefined. But OTOH, I think we need *something* at the top of the file for our devs to know what they can expect the component's state to be. Could be a docstring, but failing that, I propose we leave the empty declarations.

src/components/Map/CanadaMap.js
```diff
@@ -39,13 +40,13 @@ var CanadaMap = React.createClass({
...
             srs: "EPSG:4326",
-            colorscalerange: "-50,11.0",
+            colorscalerange: "220,320",
             logscale: false
         };
     },
     getWMSParams: function() {
         var params = {layers: this.props.dataset + "/" + this.props.variable};
-        $.extend(params, this.props);
+        _.extend(params, _.pick(this.props, 'noWrap', 'format', 'transparent', 'styles', 'time', 'numcolorbands', 'version', 'srs', 'colorscalerange', 'logscale'));
```
jameshiebert: This will work well. We can further generalize this, so that it takes a list of param names as an argument. Then we can differentiate between the leaflet initialization params (noWrap, transparent, opacity, etc.) and those that are updateable by the component (styles, time, logscale).

src/components/MapController/MapController.js
```diff
@@ -52,25 +84,42 @@ var MapController = React.createClass({
                     ]
     var color_scales = [['false', 'Linear'], ['true', 'Logarithmic']]
 
+    var ids = this.props.meta.map(function(el){return el.unique_id})
+
+    // Only render map if required paramers exist in state/props
```
jameshiebert: I don't like this. If the user is on a slower network, or there's a lag on the metadata AJAX request, then there's a giant blank white space on the page until it returns. Plus then it delays the start of the WMS tile loading until after the metadata comes back. Basically, it makes everything wait on one single request, even things that shouldn't be dependent on it.

Can't we just render the map with time, dataset and variable undefined?

src/components/MapController/MapController.js
```diff
@@ -12,37 +13,68 @@ import styles from './MapController.css';
...
 var MapController = React.createClass({
 
-    propTypes: {
-        onSetArea: React.PropTypes.func.isRequired,
-    },
+  propTypes: {
+    variable: React.PropTypes.string,
+    meta: React.PropTypes.array,
+    onSetArea: React.PropTypes.func.isRequired,
+  },
```
jameshiebert: `time` should be in the list too, right?

src/components/MapController/MapController.js
```diff
@@ -12,37 +13,68 @@ import styles from './MapController.css';
...
-      dataset: "pr-tasmax-tasmin_day_BCSD-ANUSPLIN300-CanESM2_historical-rcp26_r1i1p1_19500101-21001231",
-    }
-  },
 
   updateSelection: function(param, selection) {
     var update = {}; update[param] = selection;
     this.setState(update);
   },
 
+  updateTime: function(timeidx) {
```
jameshiebert: Good.

src/components/MapController/MapController.js
```diff
@@ -12,37 +13,68 @@ import styles from './MapController.css';
...
+
   handleSetArea: function(wkt) {
-    // TODO: Do something more here?
-    // Really just here so it's not invisibly transferred from parent props, not directly needed
     this.props.onSetArea(wkt);
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    this.setState({
+      dataset: nextProps.meta[0].unique_id,
```
jameshiebert: Rather than always selecting `meta[0]`, would it be better to tie it to the Dataset selector?

src/components/MapController/MapController.js
basilveerman: `timeidx` and `wmstime` and are contained in the state. Now documented in 0e7dea3.

src/components/Map/CanadaMap.js
basilveerman: Opened #6

src/components/AppController/AppController.js
basilveerman: Now documented in 0e7dea3.

src/components/MapController/MapController.js
basilveerman: Yes. It would be better to select the first dataset available, but we currently do not have a way to enforce default selection in the `Selector` component.

PR #7, opened 2015-12-17
Seasonal plot

mdfischer:
Updated DataGraph component to accept and seasonal plot data returned by the backend timeseries.py API call. Mainly involved changes to the local util.js, where the function parseC3Data was renamed to parseDataForC3 (still pertaining to plotting data returned by the backend data.py call), and a new function was created called parseTimeSeriesForC3 for handling this new seasonal plot functionality.
NOTE: test data is hardcoded in GraphOverlay.js for proving correct operation.


basilveerman:
Tests are not currently passing. The function names in `util` have changed and this should be updated in the test. Also, the `parseTimeSeriesForC3` formatter should have a test of it's own.

src/components/DataGraph/util.js
```diff
@@ -35,13 +40,92 @@ var parseC3Data = function(data) {
...
+    var fallSeries = [];
+    var winterSeries = [];
+    var seasonalLabel = ['Seasonal Average'];
+    var annualSeries = ['Annual Average'];
+
+    var idx = 0;
+    for (let key in graph_data['data']) {
+        var val = graph_data['data'][key].toFixed(PRECISION);
+        var timestep = moment(key, moment.ISO_8601);
+        var month = timestep.format('MMMM');
```
basilveerman: Good use of moment.

src/components/DataGraph/util.js
```diff
@@ -35,13 +40,92 @@ var parseC3Data = function(data) {
...
+            springSeries.push(val, val, val);
+        }
+        else if (idx === 14){
+            summerSeries.push(val, val, val);
+        }
+        else if (idx === 15){
+            fallSeries.push(val, val, val);
+        }
+        else if (idx === 16){
+            annualSeries.push(val, val, val, val, val, val, val, val, val, val, val, val);
```
basilveerman: This is pretty bulky... with underscore you can reduce this with something like `_.times(5, function(){return this}, val)`

src/components/DataGraph/util.js
```diff
@@ -35,13 +40,92 @@ var parseC3Data = function(data) {
...
     return [allModelsData, axisInfo];
 }
 
-export default parseC3Data
\ No newline at end of file
+
+var parseTimeSeriesForC3 = function(graph_data) {
+
+    var modelName = String(graph_data['id']);
+    var yUnits = String(graph_data['units']);
```
basilveerman: Do we need to cast these as `String`? Doesn't the data come in as string?

src/components/DataGraph/util.js
```diff
@@ -35,13 +40,92 @@ var parseC3Data = function(data) {
                     }
                     yAxisCount++;
                 }
-            }              
+            }
```
basilveerman: This probably just slipped by, but if you're fixing white space, best to fix it *all* the way.

src/components/DataGraph/util.js
```diff
@@ -35,13 +39,92 @@ var parseC3Data = function(data) {
...
+    var springSeries = [];
+    var summerSeries = [];
+    var fallSeries = [];
+    var winterSeries = [];
+    var seasonalLabel = ['Seasonal Average'];
+    var annualSeries = ['Annual Average'];
+
+    var idx = 0;
+    for (let key in graph_data['data']) {
+        var val = graph_data['data'][key].toFixed(PRECISION);
```
basilveerman: What's the purpose of rounding here? For the labels? If so, we should just round when creating the labels and leave the data intact.

PR #13, opened 2016-01-21
MapController time metadata request

### LINKED ISSUE #12: Require MapController to request time metadata of a selected dataset
### Problem
Merging pacificclimate/climate-explorer-backend#17 removes `time` from the global metadata and requires that we explicitly source the time metadata when needed. This is only when the [time of year index is converted to the wmstime parameter](https://github.com/pacificclimate/climate-explorer-frontend/search?utf8=%E2%9C%93&q=times) needed for map display.

### Solution
On `meta` change, further require the MapController to request the time metadata of a selected dataset, use this to convert to wmstime, then `setState` and re-render

basilveerman:
Fixes #12

### DIFF
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -97,7 +97,6 @@ var App = React.createClass({
           <Col lg={6}>
             <div className={styles.map}>
               <MapController
-                variable={this.state.variable_id}
                 meta = {this.getfilteredMeta()}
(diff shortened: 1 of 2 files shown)


jameshiebert:
Those two comments aside, I think it looks great and seems to perform well.

basilveerman:
Comments addressed in e54554c and df5b155.

PR #15, opened 2016-02-13
Update to Babel 6

basilveerman:


jameshiebert:
@basilveerman did you still want to consider this PR? If so, could you rebase onto master, fix the tests and maybe comment on what this will do for us?

basilveerman:
This doesn't strictly give us anything new, but *should* happen sooner or later to keep up to date. There are however unresolved issues with Babel6/React/Jest testing that will need to be ironed out before we can merge back in.

PR #16, opened 2016-02-17
Datatable export

mdfischer:


webpack.config.js
```diff
@@ -73,6 +73,16 @@ module.exports = {
         hot: true,
         inline: true,
         progress: true,
-	colors: true
-    }
+	    colors: true
+    },
+    devtool: (process.env.NODE_ENV !== 'production') ? 'source-map' : '',
```
jameshiebert: I think you had a bad merge or rebase here... this is a duplicate line/key in the dict.

webpack.config.js
```diff
@@ -73,6 +73,16 @@ module.exports = {
         hot: true,
         inline: true,
         progress: true,
-	colors: true
-    }
+	    colors: true
```
jameshiebert: Mixing tabs and spaces == bad

src/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -1,41 +1,41 @@
...
 import TestUtils from 'react-addons-test-utils';
 
 const DataTable = require('../DataTable');
 
-const testData = {
-  "tasmin_Amon_CanESM2_historical_r1i1p1_19610101-19901231": 
-  {
+const testData = [ 
+  { 
     "median": 278.34326171875,
```
jameshiebert: You're just testing for trutiness down below, so why bother with listing a bunch of data points to the 10th decimal place? It just ends up being noise. If your "test data" isn't actually required to run the test, then perhaps you should consider whether it even needs to be there? Corollary: if you test data doesn't need to be there, then what are you actually testing?

jameshiebert:
Make sure to run `eslint` and use the suggestions. There are tons of whitespace and indentation problems.

src/components/DataTable/DataTable.js
```diff
@@ -1,24 +1,23 @@
...
             <div id={'table'}>
-                <BootstrapTable data={tableData} striped={true} hover={true} >
-                    <TableHeaderColumn dataField="time" isKey={true} dataAlign="center" dataSort={true}>Model Period</TableHeaderColumn>
+                <BootstrapTable data={this.props.data} striped={true} hover={true} >
+                    <TableHeaderColumn dataField="model_period" isKey={true} dataAlign="center" dataSort={true}>Model Period</TableHeaderColumn>
                     <TableHeaderColumn dataField="run" dataAlign="center" dataSort={true}>Run</TableHeaderColumn>
                     <TableHeaderColumn dataField="min" dataAlign="center" dataSort={true}>Min</TableHeaderColumn>
                     <TableHeaderColumn dataField="max" dataAlign="center" dataSort={true}>Max</TableHeaderColumn>
-                    <TableHeaderColumn dataField="mean" dataAlign="center" dataSort={true}>W.Mean</TableHeaderColumn>
+                    <TableHeaderColumn dataField="w_mean" dataAlign="center" dataSort={true}>W.Mean</TableHeaderColumn>
```
jameshiebert: These are not weighted statistics, so you should omit the "w"s.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+      },
+      "tasmin_Amon_CanESM2_historical_r1i1p1_19710101-20001231": 
+      {
+        ...
+      }
+    };
+*/
+var parseBootstrapTableData = function(data) {
+    var flatData = [];
+    var model_count = 0;
```
jameshiebert: This is never used.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    for (let model in data) {
+        var year_range_re = new RegExp("[0-9]{8}","g");
+        var year_range = [];
+        var lastIndex = 0;
+        while (year_range_re.test(model)){
```
jameshiebert: You can eliminate this whole loop if you rewrite your regular expression to capture the info that you want to use.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+      {
+        ...
+      }
+    };
+*/
+var parseBootstrapTableData = function(data) {
+    var flatData = [];
+    var model_count = 0;
+    for (let model in data) {
+        var year_range_re = new RegExp("[0-9]{8}","g");
```
jameshiebert: You're only matching against a single line and only searching for one thing, so using the "g" (global) flag is wrong.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+        "stdev": 22.509726901403784,
+        "run": "r1i1p1"
+      },
+      "tasmin_Amon_CanESM2_historical_r1i1p1_19710101-20001231": 
+      {
+        ...
+      }
+    };
+*/
+var parseBootstrapTableData = function(data) {
```
jameshiebert: You should be able to simplify this whole function by doing something like:
~~~js
return _.map(data, function(val, key) {
...
});

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+        };
+        flatData.push(modelInfo); 
+    }
+    return flatData;
+}
+
+/*
+    Helper function for exportTableDataToWorksheet, creates summary rows that appear at the top of the exported worksheet
+*/
+var createWorksheetSummaryCells = function(summary_data, time_of_year) {
```
jameshiebert: This entire method relates only to a "summary" table, so how about making it read easier by eliminating "summary" from all of the variable names below? The reader *knows* that `num_summary_rows` is the number of rows in the summary table, because there's only one table here. So just call it `num_rows`. Likewise for the rest of the variables in the method.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    for(var R = 0; R < num_summary_rows; ++R) {
+        for(var C = 0; C < num_summary_cols; ++C) {
+            if(R == 0) cell = {v: summary_header[C]};
+            else if(R == 1) {
+                if(summary_keys[C] == 'time_of_year') {
+                  cell = {v: time_of_year};      
+                }
+                else cell = {v: summary_data[summary_keys[C]]};
+            }
+            else cell = {v: ""};
```
jameshiebert: Always use enclosing `{}` even for one line blocks. Omitting them risks bugs later on.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    // store worksheet cell contents to be later encoded as per output format
+    var cells = [];
+    var cell;
+    // write summary rows at top of worksheet
+    var summary_header = ["Model", "Emissions Scenario", "Time of Year", "Variable ID", "Variable Name"];
+    var summary_keys = ["model_id", "experiment", "time_of_year", "variable_id", "variable_name"];
+
+    var num_summary_rows = 3;
+    var num_summary_cols = summary_keys.length
+    for(var R = 0; R < num_summary_rows; ++R) {
```
jameshiebert: OK, I end up seeing "loops" like this a lot, so remember this rule of thumb: if you have a conditional inside a loop that only happens one time, it generally shouldn't be in the loop. In this case, you have a row loop that executes 3 times, but does 3 completely different things on each iteration. That being the case, there's no point to having a loop here! Just do each row separately and append them all together at the end.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    // populate worksheet with table data
+    var column_labels = ["Model Period", "Run", "Min", "Max", "W.Mean", "Median", "W.Std.Dev", "Units" ];
+    var data_keys = ["model_period", "run", "min", "max", "w_mean", "median", "w_stdev", "units" ];
+    var num_data_rows = Object.keys(data).length + 1;
+    var num_data_cols = data_keys.length
+
+    for(var R = 0; R < num_data_rows; ++R) {
+        for(var C = 0; C < num_data_cols; ++C) {
+            // create header row
+            if(R == 0) cell = {v: column_labels[C]};
```
jameshiebert: Use curly braces for the blocks.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+            // create cell object: .v is the actual data
```
jameshiebert: And... what is `.t`? Maybe some notes in the function header about what is the return value of this function, because it's not overtly apparent.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    Helper function for exportTableDataToWorksheet, combines summary rows, data column headers, and data into one worksheet
+*/
+var assembleWorksheet = function (summary_cells, data_cells) {
+    var R = 0;
+    var C = 0;
+    var cell;
+    var cell_ref;
+    var ws = {};
+
+    for(R = 0; R < summary_cells.num_rows; ++R) {
```
jameshiebert: The inside two lines for both of these sets of loops are essentially the same. Why not combine the data first and *then* loop over it only once.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    // create workbook object containing one or more worksheets
+    var wb = {}
+    wb.Sheets = {};
+    wb.SheetNames = [];
+
+    // convert timestep ID (0-16) to string format
+    var times_of_year = ["January", "February", "March", "April", "May", "June", "July", "August", "September", 
+                      "October", "November", "December", "Winter - DJF", "Spring - MAM", "Summer - JJA", 
+                      "Fall - SON", "Annual"];
+    var time_of_year = times_of_year[this.state.dataTableTimeOfYear];
```
jameshiebert: "Utility" functions are supposed to be atomic and independent of any context. I don't think that using `this` in this function is a good idea.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+
+    // prepare summary cells
+    var summary_cells = createWorksheetSummaryCells(this.props, time_of_year);
+
+    // if time_of_year is a season, overwrite it with a short version label for use in output filename (e.g. "Summer - JJA" -> "Summer")
+    var short_season_label = time_of_year.substr(0, time_of_year.indexOf(' ')); // will return string before whitespace
```
jameshiebert: `time_of_year.split(' ')[0]` is a bit more idiomatic.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    // add worksheet to workbook. Note: ws_name will be truncated to 31 chars in XLSX export to meet Excel limitation
+    var ws_name = "Stats_Table_" + this.props.variable_id + "_" + time_of_year;
+    wb.SheetNames.push(ws_name);
+    wb.Sheets[ws_name] = ws;
+
+    // Note: we will probably want to split the rest of this out into a separate function 
+    // to combine multiple worksheets if/when we want to export additional data (e.g. DataGraph points)
+    function to_csv(workbook) {
+        var result = [];
+        workbook.SheetNames.forEach(function(sheetName) {
```
jameshiebert: Use `_.map(workbook.Sheets, XLSX.utils.sheet_to_csv)` to eliminate the need for `result` and the `result.push()` line.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+                // Uncomment the following 2 lines of code to support multiple worksheets
+                // result.push("SHEET: " + sheetName);
+                // result.push("");
+                result.push(csv);
+            }
+        });
+        return result.join("\n");
+    }
+
+    function s2ab(s) {
```
jameshiebert: What does this function do? The abbreviated name does not make that obvious.

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    return ws;
+}
+
+/*
+    Takes current data displayed in the DataTable and contextual data from user input, 
+    creates an XLSX or CSV file, and serves it to the user for download.
+*/
+var exportTableDataToWorksheet = function(data, format) {
+    // create workbook object containing one or more worksheets
+    var wb = {}
```
jameshiebert: Change this to:
~~~js
var wb = {
    Sheets: {},
    SheetNames: []
};
~~~

src/core/util.js
```diff
@@ -283,4 +285,221 @@ var parseTimeSeriesForC3 = function(graph_data) {
...
+    ws['!ref'] = XLSX.utils.encode_range(range);
+    return ws;
+}
+
+/*
+    Takes current data displayed in the DataTable and contextual data from user input, 
+    creates an XLSX or CSV file, and serves it to the user for download.
+*/
+var exportTableDataToWorksheet = function(data, format) {
+    // create workbook object containing one or more worksheets
```
jameshiebert: There are multiple places in this function where the code is complicated by trying to support multiple sheets, but we only ever produce one. And CSV only supports one "sheet" anyways. So why not simplify the code and only handle one sheet since that's all we need?

jameshiebert:
OK, I've spent a considerable amount of time reviewing this PR. Overall, I can say that the code runs and works and from a user perspective, the changes are simple and effective. That said, I think that the code *quality* needs more work.

Code Climate has a [number of suggestions](https://codeclimate.com/github/pacificclimate/climate-explorer-frontend/compare/datatable-export) for how to improve the quality. Overall, this PR drops the quality by about half a point on a four point scale (so 12.5%). All of its suggestions are easy to address.

More qualitatively, I notice that in all of these utility methods, you've written them like one would
write C code. So you're not utilizing the benefits of a dynamically typed language, being able to iterate over an object itself (rather than maintaining a bunch of index values). I think you could stand to take a shot at rewriting them as more idiomatic JavaScript. It's not that they're necessarily incorrect, but they'll be harder to maintain over time, and much harder for other JavaScript programmers to understand.

I've given a number of in line suggestions for how to do this. Additionally, I think that it may be helpful for you to read over [some guidelines for writing idiomatic JavaScript](https://github.com/rwaldron/idiomatic.js/).

basilveerman:
@mdfischer Since @jameshiebert mentioned eslint, do keep in mind that it's currently set up using the Airbnb standard which includes some es6 recommendations that we don't use/enforce yet. Disregard any inappropriate suggestions as we iron out that config.

src/core/util.js
mdfischer: On line 96 of DataController.js

src/core/util.js
mdfischer: I left that in there for the future possibility that we would want to support exporting multiple datasets at once, across multiple sheets, e.g. DataTable content + DataGraph points

src/core/util.js
mdfischer: Because the number of columns in the summary rows is different than the number of columns in the data rows, and the XLSX.utils.encode_cell() function needs the row & column indices of each valid cell.

src/core/util.js
mdfischer: Most of this C-style looping code and references to .t, .v, etc. come from the examples in the docs for js-xlsx.  I will add a comment with the link: https://github.com/SheetJS/js-xlsx

src/core/util.js
mdfischer: Also from the js-xlsx example code.  I've changed the name to xml_to_binary_string() to be clearer.

src/core/util.js
mdfischer: This also came from the example code in the js-xlsx docs, but sure, we can improve it.

webpack.config.js
mdfischer: Fixed.

src/components/DataTable/__tests__/DataTable-test.js
mdfischer: Good point. It now only checks if the DataTable renders, with no test data.

src/components/DataTable/DataTable.js
mdfischer: Fixed (here and everywhere else).

src/core/util.js
mdfischer: Good catch.  Gone.

src/core/util.js
mdfischer: Done.

src/core/util.js
mdfischer: Done.

src/core/util.js
mdfischer: Done.

mdfischer:
@jameshiebert This should be ready to go.

jameshiebert:
This looks much, much better after addressing my comments. Thanks for taking the time to work on it and improve the code quality. IMO it's ready for merging with master.

PR #17, opened 2016-03-09
Feature/polygon import export

basilveerman:


jameshiebert:
Splitting out the `GeoExporter` and `GeoLoader` into components separate from the `MapController` makes this code much, much cleaner than it would be otherwise. Excellent work. I'm glad that we were able to work through the dilemma of a variable (`area`) that is being passed up and down the component stack (other implementations could have been much more confusing and error prone). In any case, the functionality seems solid, so I'll be happy to merge this.

There *are* a couple items nagging at me:
* We don't have any testing in place for these features. I'm not actually sure of what would be a productive way of testing it... it might be sufficient to simply test the load/save functions of the `g` object from `geo.js`.
* Styling on imported polygons is different from that of drawn polygons. I'll just create a separate issue for this. I don't think that it should hold back the merge.
* Like everything else in the `MapController`, the buttons continue to take up more map real estate. I think that we should deal with this all at once, when we have all of the functionality in place. Again, it shouldn't hold up the merge.

Anyways, excellent work!

basilveerman:
> Anyways, excellent work!

Likewise! I was having some difficulties sorting out how `area` propagated down/up the stack and was glad to have the help.

PR #21, opened 2016-03-09
Specify default styling of imported polygon to match leaflet-draw defaults

### LINKED ISSUE #18: Inconsistent styling for map polygons
Analysis areas can now be created on the map through two different user actions:

1. Drawing them on the map
1. Loading them from the GeoLoader

The resulting polygon's styling (fill color, line color, line width, transparency, etc.) differs between the two input methods. We should make them consistent.

basilveerman:
Fixes #18


mdfischer:
Imported polygons show up with blue outline and light shading, not the pink (#f06eaa) specified by the defaults.

PR #23, opened 2016-03-10
Fix map control positioning

basilveerman:


src/components/MapController/MapController.css
```diff
@@ -1,12 +1,10 @@
-.selector {
-}
-
 .map {
+    position: relative;
```
jameshiebert: Ahhhh, so you position the map 'relative', but leave everything at 0px. Then it's technically a "positioned" element and becomes the closest positioned ancestor of the controls. Excellent; that's what I was missing. Thanks!

PR #24, opened 2016-03-14
Add shapefile import

basilveerman:


basilveerman:
Does not currently perform error/complexity checking. Added issue #25 to address this.

src/core/geo.js
```diff
@@ -104,14 +96,41 @@ var g = {
...
+
+  load: function (file, success) {
+    /* All load functions must call `success` handler with a GeoJSON feature */
+    var ext = file.name.split('.')[1];
+
+    if (_.contains(['geojson', 'json'], ext)) {
+      this.loadTextFormat(file, success);
+    } else if (ext === 'zip') {
+      this.loadShapefile(file, success);
+    }
```
jameshiebert:
```js
else {
// inform the user about what they should do?
}
```

src/core/geo.js
```diff
@@ -104,14 +96,41 @@ var g = {
...
+      shp(evt.target.result).then(function (geojson) {
+        success(geojson);
+      });
+    };
+
+    reader.readAsArrayBuffer(file);
+  },
+
+  load: function (file, success) {
+    /* All load functions must call `success` handler with a GeoJSON feature */
```
jameshiebert: It's a *little* confusing reading the code with the success handler being passed down two levels. As long as that stack doesn't get any deeper (i.e. as long as the `load[type]file` methods don't do any more asynchronous operations) then we're probably OK. If it does get any deeper with another pass through, we may want to reorganize a bit. I think it's good for now, though.

jameshiebert:
Once you add an `else` clause in the `load()` function, I think that it's good to go.

PR #27, opened 2016-03-16
Cleanup/eslint fix

basilveerman:
Down to 224 errors from 2000+. All the remaining errors are in `util`.

Functionally, nothing should have changed, but since this is a pretty wide swath of destruction, an extra set of eyes would be helpful. Tests pass and no operational errors appear for me.


jameshiebert:
Pfew, *that* was not a fun PR to review! But it probably wasn't very fun to fix either, so thanks for doing this. I didn't see anything obviously wrong and found the same experience (tests pass, and everything in the UI appears to work). I'll go ahead and merge this.

To avoid the creeping death of having to do this again down the road, should we install a pre-commit hook such as this [this one](https://github.com/jhurliman/precommit-hook)?

basilveerman:
@jameshiebert @mdfischer  Check out [this](https://medium.com/@dan_abramov/lint-like-it-s-2015-6987d44c5b48#.qsme7etll) article. Most of that setup has already been done, but it also goes through the process of setting up automatic linting in sublime which is *really really* helpful.

Yes, I agree with a pre-commit hook. I've never written one before but using an entire package seems like overkill when all it needs to do is run `npm run lint` and check the return code. I'll look into it.

We also have the option of using an eslint loader for webpack that will enforce the rules before building the bundle, but I tried that out and found it was too aggressive.

Thanks for the speedy review.

PR #30, opened 2016-03-30
Merge MoTI specific controllers/options

basilveerman:


basilveerman:
@jameshiebert To clarify, the idea is to add react-router at a later date and have multiple Apps under different routes.  For that to happen, we will need to change some of the configuration varibles and ideally eliminate some duplicated code with composed components/mixins.

PR #38, opened 2016-04-05
Feature/map colorbar

### LINKED ISSUE #32: Add Colorbar to map pane
Map symbology is currently quite meaningless with no reference. We should add a colorbar to the map viewer.

basilveerman:
Add a Leaflet.Control extension to display a colobar for any ncWMS layer. Binds to supplied layer's 'loading' event to automatically update when the layer is updated/changed. Compatible with any ncWMS server which fulfills `minmax` and `layerDetail` `GetMetadata` requests. Respects manually set layer min/max, else uses dataset defaults.

Fixes #32

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -33,5 +33,6 @@
     "react/sort-comp": 1,
     "prefer-template": 0,
     "jsx-quotes": [2, "prefer-single"],
+    "no-param-reassign": [2, { "props": false }]
   }
 }
--- a/src/components/Map/CanadaMap.js
+++ b/src/components/Map/CanadaMap.js
@@ -2,7 +2,6 @@ import React from 'react';
 import _ from 'underscore';
 import leafletImage from 'leaflet-image';
 import { saveAs } from 'filesaver.js';
-import cn from 'classnames';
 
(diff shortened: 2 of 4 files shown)


jameshiebert:
So I have two main comments:

1. It looks great, works great and does the job. However, we may want to future-proof it a little better. Should we convert this to a proper React component? Have its parameters actually bound to props/state? It's not overly obvious to a casual reader which are the important properties that affect an updated and what gets updated (yes, I realize that this is a product of how I wrote it initially in the data portal, but maybe now is a good time to fix that :) ).
1. You've done a good job of specifying the dependencies for ncWMS in the PR, but it would be good to also put that in comments at the top of the source file.

basilveerman:
Regarding the second point: I've added some documentation at the top.

As for the first, I don't think that writing it as a react component would be beneficial. The idea of adding the control and passing it only the layer that it represents is easy. The functionality is entirely abstracted and we do not need to maintain it within the map component. If it were to be a react component, we would need to add additional complexity (event handling, element positioning, passing props) within the CanadaMap component that doesn't give us any functional advantage.

PR #39, opened 2016-04-05
Add ncWMSAutoscaleControl and add it to the map

### LINKED ISSUE #33: Add additional map symbology options
Map symbology currently defaults to the entire data range of the entire spatial and temporal extent. This often creates washed out maps when viewing a localized region with minimal spatial variability in relation to the entire range.

We should consider adding additional options for controlling map symbology like:
- Manually set min/max
- Autoset min/max to viewport

basilveerman:
Adds a Leaflet control to automatically set an ncWMS layer colorscalerange WMS parameter to the min/max of the current map view.

(Partially) Fixes #33.

### DIFF
--- a/src/core/leaflet-ncwms-autoset-colorscale.js
+++ b/src/core/leaflet-ncwms-autoset-colorscale.js
@@ -22,6 +22,7 @@ var ncWMSAutoscaleControl = L.Control.extend({
 
     this.button = L.DomUtil.create('a', '', this.container);
     this.button.innerHTML = 'AS';
+    this.button.style.fontWeight = 'bold';
 
     // Set up event handling
     L.DomEvent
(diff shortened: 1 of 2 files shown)


src/core/leaflet-ncwms-autoset-colorscale.js
```diff
@@ -0,0 +1,59 @@
+/*
+ * Relies upon global Leaflet variable 'L'
+ * Compatible with any ncWMS server which fulfills the `minmax`
+ * `GetMetadata` requests.
+*/
+
+var ncWMSAutoscaleControl = L.Control.extend({
+  options: {
+    position: 'topleft',
```
jameshiebert: Can we group this in the `bottomright` with the scale bar?

jameshiebert:
Aside from the positioning of the button, everything looks good to me.

Issue #43, opened 2017-05-11
Robustify climate data plots


jameshiebert:
The climate explorer frontend provides multiple plot types to explore various properties of climate data. These include:

* Annual cycle plot
  * generally shows seasonal variation
  * requires monthly means, seasonal means and annual means (17 timesteps)
* Projected change plot
  * generally shows a temporally coarse (e.g. 30 year means) line plot
  * requires climatological means
* Timeseries plot
  * generally shows more temporal detail
  * requires annual means or greater

Not all datasets that are exposed by the climate explorer backend will necessarily have the values available for the type of plot that is being displayed. For example, an annual timeseries dataset will not have the seasonal values required for the Annual Cycle plot.

At this point, none of the plots communicate with the user as to when they are loading data or when they fail to load data. Which makes things suck for the user and also makes it really confusing for us when we're developing.

For this ticket we should review the ``DataControllerMixin`` component, particularly the methods ``getDataPromise()``, ``getStatsPromise()`` and ``getTimeseriesPromise()``, and look for ways to add error handling to the promise. Presumably, we'd set a loading icon for the component when the promise is created and then set an error graphic or "Data not available message" in the `.error()` handler.

PR #44, opened 2017-05-17
Add some basic error handling

### LINKED ISSUE #45: Don't use jQuery for Ajax
We're importing jQuery just to use its Ajax facility. (There are 12 references in code to the jQuery alias `$`, and they're all `$.ajax()` calls.)

Using jQuery for this purpose seems a little perverse:

1. jQuery is not mainly about dispatching HTTP requests. That's a lot of overhead to get an HTTP package.
2. The promise-like objects jQuery returns [are not standards-compliant promises](https://blog.domenic.me/youre-missing-the-point-of-promises/), and the differences are [confusing and troublesome](http://stackoverflow.com/questions/23744612/problems-inherent-to-jquery-deferred-jquery-1-x-2-x). And we're using these promisey objects, and I think [already hitting trouble](https://github.com/pacificclimate/climate-explorer-frontend/pull/44#discussion_r117149437).
3. [There are HTTP libraries that return standards-compliant promises](https://daveceddia.com/ajax-requests-in-react/), `axios` and `fetch` (itself likely to become a standard) for two.

If we want to use promises for async operations (and I think it's a good idea, though [not for all applications](http://sealedabstract.com/code/broken-promises/)), may I suggest we retool away from jQuery and its icky promisey things before we have much more code written around it? A quick perusal of the CE frontend code suggests this won't be hard, and it could save us some headaches later.

corviday:
Addresses #43 and fixes #45

### DIFF
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import urljoin from 'url-join';
+import axios from 'axios';
 
 var AppMixin = {
   getInitialState: function () {
@@ -11,30 +12,30 @@ var AppMixin = {
   componentDidMount: function () {
     var models = [];
     var vars;
-    $.ajax({
-      url: urljoin(CE_BACKEND_URL, 'multimeta'),
-      data: { ensemble_name: CE_ENSEMBLE_NAME },
-      crossDomain: true,
-    }).done(function (data) {
-      for (var key in data) {
-        vars = Object.keys(data[key].variables);
+       
+    axios({
+      baseURL: urljoin(CE_BACKEND_URL, 'multimeta'),
+      params: { ensemble_name: CE_ENSEMBLE_NAME },
+      }).then(function (response) {
+        for (var key in response.data) {
+          vars = Object.keys(response.data[key].variables);
 
-        for (var v in vars) {
-          models.push(_.extend({
-            unique_id: key,
-            variable_id: vars[v],
-            variable_name: data[key].variables[vars[v]],
-          }, _.omit(data[key], 'variables')));
-        }
-      }
+          for (var v in vars) {
+            models.push(_.extend({
+              unique_id: key,
+              variable_id: vars[v],
+              variable_name: response.data[key].variables[vars[v]],
+              }, _.omit(response.data[key], 'variables')));
+            }
+          }
 
-      this.setState({
-        meta: models,
-        model_id: models[0].model_id,
-        variable_id: models[0].variable_id,
-        experiment: models[0].experiment,
-      });
-    }.bind(this));
+        this.setState({
+          meta: models,
+          model_id: models[0].model_id,
+          variable_id: models[0].variable_id,
+          experiment: models[0].experiment,
+          });
+        }.bind(this));
   },
 
   handleSetArea: function (wkt) {
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -16,7 +16,7 @@ var AppMixin = {
     axios({
       baseURL: urljoin(CE_BACKEND_URL, 'multimeta'),
       params: { ensemble_name: CE_ENSEMBLE_NAME },
-      }).then(function (response) {
+      }).then(response => {
         for (var key in response.data) {
           vars = Object.keys(response.data[key].variables);
 
@@ -28,14 +28,14 @@ var AppMixin = {
               }, _.omit(response.data[key], 'variables')));
             }
           }
-
-        this.setState({
-          meta: models,
-          model_id: models[0].model_id,
-          variable_id: models[0].variable_id,
-          experiment: models[0].experiment,
-          });
-        }.bind(this));
+        
+         this.setState({
+         meta: models,
+         model_id: models[0].model_id,
+         variable_id: models[0].variable_id,
+         experiment: models[0].experiment,
+         });
+        });
   },
 
   handleSetArea: function (wkt) {
--- a/src/components/DataControllerMixin/DataControllerMixin.js
+++ b/src/components/DataControllerMixin/DataControllerMixin.js
@@ -85,17 +85,13 @@ var ModalMixin = {
       throw new Error("Error: statistical data unavailable for this model");
     }
     for(var file in response.data) {
-      if(!('mean' in response.data[file])|| (isNaN(response.data[file].mean)) ||
-        !('stdev' in response.data[file]) || (isNaN(response.data[file].stdev)) ||
-        !('min' in response.data[file]) || (isNaN(response.data[file].min)) ||
-        !('max' in response.data[file]) || (isNaN(response.data[file].max)) ||
-        !('median' in response.data[file]) || (isNaN(response.data[file].median)) ||
-        !('ncells' in response.data[file]) || (isNaN(response.data[file].ncells)) ||
-        !('units' in response.data[file]) ||
-        !('time' in response.data[file])) {
-        throw new Error("Error: statistical data for this model is incomplete"); 
-        }
+      if(_.some('mean stdev min max median ncells'.split(''),
+          attr => !(attr in respone.data[file]) || isNAN(response.data[file][attr])) ||
+          _.some('units time'.split(' '),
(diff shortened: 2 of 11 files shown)


src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -51,6 +51,16 @@ var ModalMixin = {
...
+    }).then(function(data, textstatus, jqXHR) {
+      if($.isEmptyObject(data)) {
+        return $.Deferred().reject(jqXHR, "empty", textstatus);
+      }
+      for(var run in data) {
+        if(!('data' in data[run]) || !('units' in data[run])){
+          return $.Deferred().reject(jqXHR, "incomplete", textstatus);            
+        }
+      }
+      return $.Deferred().resolve(data, textstatus, jqXHR);
```
rod-glover: I fear that this, and similar code for the other `getXPromise` functions may not do what you'd like it to do. [This is because jQuery < 3 doesn't actually implement what most of the JS world means by Promise](http://stackoverflow.com/questions/23744612/problems-inherent-to-jquery-deferred-jquery-1-x-2-x). Domenic Denicola wrote an [explanatory article](https://blog.domenic.me/youre-missing-the-point-of-promises/) about this several years ago which is still valid. 

(In my first big outing with jQuery promises about 5 years ago, I tripped all over these issues without understanding them at the time. They're gnarly, not least because the terminology overlaps but means similar but also importantly different things in the two contexts. Since then I've used and come to greatly appreciate Promises/A+ compliant promises, which are now part of the latest JavaScript standard (ES6).)

Here is the problem I believe you have:

1. `$.ajax()` returns a [`jqXHR`](http://api.jquery.com/jQuery.ajax/#jqXHR) object. 
2. A `jqXHR` object implements the [jQuery Promise](http://api.jquery.com/deferred.promise/) interface.
3. The jQuery Promise interface is a subset of the [jQuery Deferred](http://api.jquery.com/category/deferred-object/) interface.
4. The jQuery Deferred/Promise interface only provides for attaching callbacks for various Deferred events. It doesn't return the value (or, more properly a promise for the value) of what the handlers return. This is the issue taken up in the two links above.
5. This callback handler's purpose appears to be to: 
    - transform the result of the response to the Ajax request to a Deferred -- this it does
    - return that Deferred to the caller of the `getDataPromise` function -- this it doesn't, because of how jQuery Deferreds/Promises work
6. Instead, I believe  `getDataPromise` will return the original jqXHR object, and your client code will simply attach another callback to that object, not to the nice Deferreds you manufactured in the callback handler immediately above.

I could be wrong about this -- haven't used jQuery for some while -- but if this code doesn't work, the above is the reason why.

rod-glover: @corviday, James has asked me to act on my [issue/recommendation about promises](https://github.com/pacificclimate/climate-explorer-frontend/issues/45). This will cause some changes (mainly simplifying) to your code here. Let's discuss.

corviday:
As far as I can tell, this code does actually _work_. For example, I can add a line to `dataControllerMixin.getDataPromise()` that rejects the  promise every time:

```javascript
 getDataPromise: function (props, timeidx) {
    return $.ajax({
      url: urljoin(CE_BACKEND_URL, 'data'),
      crossDomain: true,
      data: {
        ensemble_name: CE_ENSEMBLE_NAME,
        model: props.model_id,
        variable: props.variable_id,
        emission: props.experiment,
        area: props.area || null,
        time: timeidx,
      },
    }).then(function(data, textstatus, jqXHR) {
      if($.isEmptyObject(data)) {
        return $.Deferred().reject(jqXHR, "empty", textstatus);
      }
      for(var run in data) {
        if(!('data' in data[run]) || !('units' in data[run])){
          return $.Deferred().reject(jqXHR, "incomplete", textstatus);            
        }
      }
      // reject for testing purposes
      return $.Deferred().reject(jqXHR, "testing promise chaining", textstatus);
      
      return $.Deferred().resolve(data, textstatus, jqXHR);
    });
  },
```

And the rejected promise will be properly consumed by `dataController.getData()` and the error message reflected to the user in the UI:

screenshot from 2017-05-18 10-14-29

But it's also not _really_ chaining promises; the second promise is directly returned in place of the first one, which I don't think is necessarily how promises are supposed to work and probably a little brittle. I tried to minimize brittleness by having the second promise use the same arguments as a rejected `$.ajax()`. Switching to a library with proper promise chaining is probably the smarter way to go, long term.

src/components/AppMixin/AppMixin.js
```diff
@@ -11,30 +12,30 @@ var AppMixin = {
...
-        experiment: models[0].experiment,
-      });
-    }.bind(this));
+        this.setState({
+          meta: models,
+          model_id: models[0].model_id,
+          variable_id: models[0].variable_id,
+          experiment: models[0].experiment,
+          });
+        }.bind(this));
```
rod-glover: Since we are using Babel, we have ES6 syntax available. It's easier to read and write the arrow syntax to create a function bound to the local `this`, as follows:

```
axios({
   ...
}).then(response => {
   for (var key in response.data)
      ...
});
```

rod-glover: This applies throughout this code.

corviday:
Test docker here:
- [climo](http://docker2.pcic.uvic.ca:8021/climo)
- [moti](http://docker2.pcic.uvic.ca:8021/moti)

src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -40,47 +42,102 @@ var ModalMixin = {
...
+    }
+    for(var file in response.data) {
+      if(!('mean' in response.data[file])|| (isNaN(response.data[file].mean)) ||
+        !('stdev' in response.data[file]) || (isNaN(response.data[file].stdev)) ||
+        !('min' in response.data[file]) || (isNaN(response.data[file].min)) ||
+        !('max' in response.data[file]) || (isNaN(response.data[file].max)) ||
+        !('median' in response.data[file]) || (isNaN(response.data[file].median)) ||
+        !('ncells' in response.data[file]) || (isNaN(response.data[file].ncells)) ||
+        !('units' in response.data[file]) ||
+        !('time' in response.data[file])) {
```
rod-glover: Could make this shorter and clearer:

```js
if (_.some('mean stdev min max median ncells'.split(' '), 
           attr => !(attr in response.data[file]) || isNaN(response.data[file][attr])) ||
   _.some('ncells units time'.split(' '), 
           attr => !(attr in response.data[file])) {
```

src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -40,47 +42,102 @@ var ModalMixin = {
...
+    return response;
+    },
+    
+    validateTimeSeriesData: function(response) {
+      if(_.isEmpty(response.data)) {
+        throw new Error("Error: timeseries data unavailable for this model.");
+      }
+      if(!('id' in response.data) || 
+          !('units' in response.data) || 
+          !('data' in response.data)) {
```
rod-glover: Could make shorter and less repetitive with:

```js
if (!_.every('id units data'.split(' '), attr => attr in response.data) { 
```

src/core/leaflet-ncwms-autoset-colorscale.js
```diff
@@ -51,8 +50,8 @@ var ncWMSAutoscaleControl = L.Control.extend({
         width: 100,
         height: 100,
       },
-    }).done(function (data) {
-      this.layer.setParams({ colorscalerange: data.min + ',' + data.max });
+    }).then(response => {
+      this.layer.setParams({ colorscalerange: response.data.min + ',' + response.data.max });
```
rod-glover: Consider using [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals). I find them easier to read and write.

jameshiebert:
One last comment on this PR... if it fixes #45 are we safe to remove jQuery from [package.json](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/package.json#L37)?

jameshiebert:
Cool, looks great. I'm going to go ahead and merge this. Thanks for following up on all of our (sometimes contradictory) comments ;)

Issue #37, opened 2016-04-01
Default value in Variable Selection drop-down is not in agreement with default variable in plot on page load.


mdfischer:
The default value shown in the Variable Selection drop-down menu is the first one in the list (alphabetically), which is different from the variable for which data is being displayed in the DataGraph.  Additionally, clicking on that same default variable in the Variable Selection drop-down menu does not trigger an update of the DataGraph to display that variable.

PR #47, opened 2017-05-26
Set selector value based on controller state

corviday:
Addresses #37.

### DIFF
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -10,10 +10,7 @@ var App = React.createClass({
 
   /**
    * Initial state set upon metadata returning in {@link App#componentDidMount}.
-   * Includes:
-   *   - model_id
-   *   - variable_id
-   *   - experiment
+   * Includes: - model_id - variable_id - experiment
    */
 
   mixins: [AppMixin],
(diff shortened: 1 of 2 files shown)


corviday:
This is a minor change, my main concern is whether setting selected values this way would have any weird side effects I didn't turn up in testing.

src/components/AppController/AppController.js
```diff
@@ -23,13 +23,13 @@ var App = React.createClass({
       <Grid fluid>
         <Row>
           <Col lg={4} md={4}>
-            <Selector label={"Model Selection"} onChange={this.updateSelection.bind(this, 'model_id')} items={this.getMetadataItems('model_id')}/>
+            <Selector label={"Model Selection"} onChange={this.updateSelection.bind(this, 'model_id')} items={this.getMetadataItems('model_id')} value={this.state.model_id ? this.state.model_id : null}/>
```
rod-glover: `this.state.model_id ? this.state.model_id : null` is more compactly expressed as `this.state.model_id || null`

Deeper questions: Do you really need to convert a falsy value to `null`? And note that it would convert a model id of `0` to `null`, also, which might not be desirable. Depends on type and range of values for that variable.

corviday: Hm, looks like I don't. `Selector` itself converts falsy `props.value` inputs to `undefined`, no need to repeat that.

Which does means we can't display a `model_id` of 0, but `model_id`s are strings, so something has already gone pretty badly if it is being attempted.

Issue #35, opened 2016-04-01
Add map title


basilveerman:
The map options pane allows a user to select different time periods and potentially different datasets. When viewing the map however, these selections are not readily apparent.  We should consider adding a map title bar/footer element detailing the specific dataset/time period being viewed.

Issue #34, opened 2016-04-01
Add timeseries data export


basilveerman:
Similar to the ability to export the DataTable, we should add the ability to export the the timeseries data in the current chart view.

PR #49, opened 2017-06-05
Add map footer showing timestamp and dataset

corviday:
Addresses #35

### DIFF
--- a/src/components/MapController/MapController.js
+++ b/src/components/MapController/MapController.js
@@ -155,6 +155,11 @@ var MapController = React.createClass({
       var year = timestamp.getFullYear();
       var runMetadata = this.props.meta.find(match => {return match.unique_id === this.state.dataset})
       var run = runMetadata.ensemble_member;
+
+      // FIXME: Time period should be determined from the metadata API
+      // which currently doesn't give time bounds information. See here:
+      // https://github.com/pacificclimate/climate-explorer-backend/issues/44
+      // When that issue is fixed, this code needs to be updated
       mapFooter = (
         <h5>
           Dataset: {run} &nbsp;
--- a/src/components/MapController/MapController.js
+++ b/src/components/MapController/MapController.js
@@ -24,7 +24,7 @@ var MapController = React.createClass({
 
   mixins: [ModalMixin],
 
-  /**
+  /*
    * State items also set from meta object array Includes: - dataset - wmstime -
    * variable
    */
@@ -138,7 +138,7 @@ var MapController = React.createClass({
       return [k, v];
     });
 
-    var map;
+    var map, mapFooter;
     if (this.state.dataset) {
       map = (
         <CanadaMap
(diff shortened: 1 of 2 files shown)


jameshiebert:
Can you rebase this?

corviday:
Test docker for this branch: 
- [Climo](http://142.104.230.42:8301/climo) 
- [Moti](http://142.104.230.42:8301/moti)

(Known issue: the stats table in the test docker is broken due to #53 )

src/components/MapController/MapController.js
```diff
@@ -24,7 +24,7 @@ var MapController = React.createClass({
 
   mixins: [ModalMixin],
 
-  /**
+  /*
```
jameshiebert: Good catch!

src/components/MapController/MapController.js
```diff
@@ -150,9 +150,23 @@ var MapController = React.createClass({
           onSetArea={this.handleSetArea}
           area={this.state.area}
         />
+      );    
+      var dataAttributes = this.state.dataset.split('_');
```
jameshiebert: We *really* don't want to use the dataset ID to extract metadata. This builds weird assumptions into our code that shouldn't be there (and means that the rules around dataset ID are too rigid, and *have* to be followed, else break the application). Dataset ID should be assumed to contain nothing more than a unique sequence of characters... it could just as well be a monotonically increasing number or a hexadecimal hash.

With that said, could you refactor this to use the information from the `medatadata` and `multimeta` API calls instead? The information should have already been retrieved by the front-end by this point in the code.

src/components/MapController/MapController.js
```diff
@@ -150,9 +150,23 @@ var MapController = React.createClass({
...
+      var run = dataAttributes[4];
+      var startYear = dataAttributes[5].slice(0, 4);
+      var endYear = dataAttributes[5].slice(9, 13);
+      var time = this.state.wmstime.slice(0,10);
```
jameshiebert: Definitely good to display the time that we're looking at, but why ``slice(0,10)``? Will that be robust against every time that might be seen?

src/components/MapController/MapController.js
```diff
@@ -170,7 +184,9 @@ var MapController = React.createClass({
                   <GeoLoader onLoadArea={this.handleSetArea} title='Import polygon' />
                 </ButtonGroup>
               </div>
-
+              <div className={styles.footer}>
```
jameshiebert: Styling looks good.

src/components/MapController/MapController.js
```diff
@@ -150,9 +150,23 @@ var MapController = React.createClass({
           onSetArea={this.handleSetArea}
           area={this.state.area}
         />
+      );    
+      var dataAttributes = this.state.dataset.split('_');
+      var run = dataAttributes[4];
```
jameshiebert: This is more of a discussion question than a comment, but is it valuable to have the variable shown on the footer? I guess that's something that the user will have selected (and will be shown above), but is there value to having it be explicitly shown on the map as well? (For future map printing, or for screenshots?)

corviday:
I've run into this several times and I'm sure I'm overlooking _something_, but how would one get the climatology timeline information (like 2010 - 2039)  from the API? As far as I know, none of the API calls provide this information. Everywhere in the codebase that displays the climatology period to the user, as far as I know, determines it by slicing the dataset id string.

I could derive the climatology period from timestamps on individual values, assuming for example that  value timestamped 2025 indicates that the climatology period is 2010-2039.  Would that be safe? Or, at least _safer_ than relying on dataset ids? :)

jameshiebert:
> but how would one get the climatology timeline information (like 2010 - 2039) from the API?

Ahhh, good point! You're right, that's not in the API, but it should be. I'll write up an enhancement issue for that in the backend, and if you want to go ahead and use the timestamp +/- 15 years, that will at least be an improvement from and safer than relying on dataset ids.

corviday:
Great, thanks. I'll switch methods, here and on Issue #34 where it also came up.

jameshiebert:
Looks good. Thanks!

PR #52, opened 2017-06-07
Experimental Pull Request, please ignore

corviday:
This pull request is intended to track down a test failure I can't reproduce on my work station. No review or merge action is needed.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -13,6 +13,8 @@ var utcDateToYYYYMMDD = function (d) {
 
 var timeseriesToC3 = function (data) {
   var c3Data;
+  var aNumber = 3.44444;
+  var fixed = aNumber.toFixed(2);
 
   c3Data = {
     x: 'x',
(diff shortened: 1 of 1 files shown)


corviday:
So, it seems like any use of the javascript builtin `Number.toFixed()` in _this particular file_ causes TravisCI to fail a test. It will fail the first test run, no matter which test that is. All subsequent ones will pass.

Other files in this repository are able to use `Number.toFixed()`. For example, it's used in `util.js` for the same purpose - formatting graph ticks.

The tests do _not_ fail when run on my workstation after executing the same commands in the same order, using the same versions of nvm, npm, and node.js as TravisCI.

I _really_ don't understand what's going on here, but it looks like the solution is just to not use `Number.toFixed()` in this particular file.

It also looks like I am not the first person to discover this particular bit of weirdness - the tooltip attribute below uses `Math.round()` with multiplying and dividing by 100 to format numbers  instead of the more straightforward `Number.toFixed()` as well.

PR #57, opened 2017-06-13
Enable downloading graph data. Addresses #34

corviday:
[Test docker here](http://docker2.pcic.uvic.ca:8303/climo).

Not thrilled with the UI and would like feedback. There are two UI flows for downloading exported data already present in the portal:

- You download map data by clicking on a small icon in the upper corner of the map, which brings up a popup to select download format
- You download statistical data by clicking on one of two buttons that say either "XLSX" or "CSV" below the stats table to select a format.

Downloading graph data seemed more similar to downloading statistical data than downloading a map, so I went with the text button approach, but I'm not thrilled with the way it looks and would be delighted to hear a better suggestion. Maybe a little icon in the upper corner of the graph, similar to the map download process, would be better?

**Known issues**: Time periods in downloaded data may not match time periods displayed on screen. Some of our thirty year datasets (most future projections) have mean data timestamped to the fifteenth year, some (most historical projections) to the sixteenth. Currently functions that generate data for download assume the former. [Eventually the API will provide this information directly and the frontend won't have to guess.](https://github.com/pacificclimate/climate-explorer-backend/issues/44)


src/components/DataController/DataController.js
```diff
@@ -172,6 +176,14 @@ var DataController = React.createClass({
               <Col lg={4} lgPush={8} md={6} mdPush={6} sm={6} smPush={6}>
                 <Selector label={"Dataset"} onChange={this.updateAnnCycleDataset} items={ids} />
               </Col>
+              <Col lg={4} lgPush={1} md={6} mdPush={1} sm={6} smPush={1}>
+                <div>
+                  Download Data
```
jameshiebert: Maybe this should be a [`ControlLabel`](https://react-bootstrap.github.io/components.html#forms-props-control-label) to give it the same styling as the other label elements on the page?

src/components/DataController/DataController.js
```diff
@@ -180,6 +192,14 @@ var DataController = React.createClass({
               <Col lg={4} lgPush={8} md={6} mdPush={6} sm={6} smPush={6}>
                 <TimeOfYearSelector onChange={this.updateProjChangeTimeOfYear} />
               </Col>
+              <Col>
+                <div>
+                  Download Data
```
jameshiebert: Likewise.

src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -27,7 +28,32 @@ var ModalMixin = {
   },
 
   exportDataTable: function (format) {
-    exportTableDataToWorksheet(this.props, this.state.statsData, format, this.state.dataTableTimeOfYear);
+    exportDataToWorksheet("stats", this.props, this.state.statsData, format, this.state.dataTableTimeOfYear);
+  },
+
+  exportTimeSeries: function(format) {
+    var runs = _.map(this.props.meta, function(run) {
+      return this.getTimeseriesPromise(this.props, run.unique_id);
```
jameshiebert: Am I reading this correctly in that when the user clicks the download button, the UI will request the timeseries from the backend (again)? On one hand, this is somewhat redundant since the UI will have already downloaded this data to present it in the plot. On the other hand, the size is tiny, so re-downloading it isn't a huge deal. Just want to make sure that we're considering that. *Is* it easy to find this data in the existing state somewhere?

src/core/__tests__/export-data.js
```diff
@@ -0,0 +1,621 @@
+var metadataForExport = {
```
jameshiebert: Where did all these data come from?

In general, I'm not a fan of the testing method of committing a 500+ line "good" dataset and then testing against it. If one of the tests stops working at some point, then rather than being able to look at the test and see a clear mismatch, we'll have to go digging through the (supposedly) good dataset and try and determine why it's not good any more.

It might be better than *not* testing, but it also might end up being harder to work with in the future.

src/core/__tests__/util-test.js
```diff
@@ -283,27 +286,83 @@ const worksheetDataRowsExpected = [
...
+  it('Assembles an exportable worksheet from summary and data cells', function () {
+    var exportfiles = require('../export');
+    var summaryCells = exportfiles.createWorksheetSummaryCells(worksheetSummaryData, worksheetTimeOfYear);
+    var dataCells = exportfiles.fillWorksheetDataCells(worksheetTestData);
+    var ws = exportfiles.assembleWorksheet(summaryCells.concat([[]], dataCells));
     expect(ws['!ref']).toEqual(worksheetRange);
   });
 });
+
+describe('createWorkSheetSummaryCells2', function () {
```
jameshiebert: What does the "2" suffix mean? Could you use a more descriptive name?

src/core/export.js
```diff
@@ -0,0 +1,303 @@
...
+
+  // assemble the worksheet and add it to the workbook.
+  // Note: sheetName will be truncated to 31 chars in
+  // XLSX export to meet Excel limitation
+  var ws = assembleWorksheet(summaryCells.concat([[]], dataCells));
+  var sheetName = `${datatype}_${metadata.variable_id}`;
+  wb.SheetNames.push(sheetName);
+  wb.Sheets[sheetName] = ws;
+
+  function xml_to_binary_string(s) {
```
jameshiebert: What's going on here? Or more specifically, why do we need to do this? (Not that I doubt that we need to, but a short comment of the "why" would be sufficiently helpful).

src/core/export.js
```diff
@@ -0,0 +1,303 @@
...
+    }
+  else if (format === 'xlsx') {
+    var wbout = XLSX.write(wb, { bookType:'xlsx', bookSST:false, type: 'binary'});
+    out_data = new Blob(
+        [xml_to_binary_string(wbout)],
+        { type:'' }
+        );
+  }
+  // serve file for download
+  filesaver.saveAs(out_data, outputFilename);
```
jameshiebert: Cool, this works great!

src/core/__tests__/export-data.js
corviday: The data is was generated by temporarily adding code to the functions in question to write their received inputs and outputs to file, so ultimately it's from the current API and netcdf files. I wasn't entirely satisfied with this approach, but wanted us to have _some_ tests. 

I would be happy to remove them, or follow suggestions on writing them differently.

src/core/__tests__/util-test.js
corviday: All the tests that climate explorer currently has test a single function (providing an input and testing it against an expected output) and are given the same name as the function they test; I followed that convention when adding tests. There was already a test of the createWorkSheetSummaryCells function, named createWorkSheetSummaryCells.

I felt that existing test was insufficient. (Specifically, the test data used in that test, worksheetSummaryData, contains a "variable_name" attribute that isn't present in the real API data and has to be fetched separately with a metadata call. So while createWorksheetSummaryCells passed its test, exported datafiles have been giving variable name as "undefined" unnoticed for a while. This is, of course, an example of the perils of using "good" datasets, as you mentioned above.)

So I thought we needed an additional test of the createWorkSheetSummaryCells functions, and named it createWorkSheetSummaryCells2, since it also tests the createWorkSheetSummaryCells function.

corviday:
I had to upgrade the `react-bootstrap` version to 0.29 to use `ControlLabel`. Couldn't go all the way to the current version of `react-bootstrap` without breaking `Selector`. It's probably an easy fix and we might have reason to want a newer `react-bootstrap` someday, but it doesn't seem very high priority right now, so I left it at 0.29.

corviday:
Changes made:

1. Instead of being collected by fresh API call(s) to the backend, data for export is now parsed from the C3 graph configuration objects from the `DataController` state. This has advantages:
- We don't need to make more API calls
- Everything is a lot simpler, we just need one function to parse downloadable data from a graph instead of a separate function to recreate the data for each type of graph it's possible to show
- It will be trivially easy to make it possible to download data from another type of graph (such as the upcoming compare-two-variables graph)

A possible disadvantage is that a user can no longer download data from all ensemble runs at once, only runs actively shown on the graph right then. I don't anticipate that being much of an issue, since this tool is intended for engineers who are looking more for means and general trends than individual runs.  So this seems like a reasonable tradeoff until we hear otherwise from users. 

2. Used `ControlLabel` to format download labels to match the rest of the page.

3. Reworked the tests, made them match the download-a-graph functionality and much simpler test data (fewer functions to test and smaller inputs for each of them).

I'd appreciate you guys taking a look when it's convenient, getting this merged will make my life easier on the next project.

src/components/DataController/DataController.js
```diff
@@ -172,6 +177,14 @@ var DataController = React.createClass({
               <Col lg={4} lgPush={8} md={6} mdPush={6} sm={6} smPush={6}>
                 <Selector label={"Dataset"} onChange={this.updateAnnCycleDataset} items={ids} />
               </Col>
+              <Col lg={4} lgPush={1} md={6} mdPush={1} sm={6} smPush={1}>
+                <div>
+                  <ControlLabel>Download Data</ControlLabel>
+                  <br/>
```
rod-glover: `<br/>` is generally considered undesirable markup. It's not semantic and it forces layout information into what should be content specification only (IOW, it doesn't say what, it says how). If two things need a break between them, there are some options. The simplest ones in this case would be one of the following:

- Put `ControlLabel` and/or the `Button` elements into their own `div`s . That is more properly semantic, though you are relying on the standard styling of div to mean "vertical layout".
- In the CSS, select this ControlLabel (adding a class to it if needed) and style it `display: block`.

src/components/DataController/DataController.js
```diff
@@ -180,6 +193,14 @@ var DataController = React.createClass({
               <Col lg={4} lgPush={8} md={6} mdPush={6} sm={6} smPush={6}>
                 <TimeOfYearSelector onChange={this.updateProjChangeTimeOfYear} />
               </Col>
+              <Col>
+                <div>
+                  <ControlLabel>Download Data</ControlLabel>
+                  <br/>
```
rod-glover: As above. If you go the CSS route, your selector should capture all these cases. There may even be an existing Bootstrap class you can apply.

src/components/DataController/DataController.js
rod-glover: See next comment also.

src/components/MotiDataController/MotiDataController.js
```diff
@@ -83,6 +83,11 @@ var MotiDataController = React.createClass({
     return (
       <div>
         <h3>{this.props.model_id + ' ' + this.props.variable_id + ' ' + this.props.experiment}</h3>
+        <div>
+          <ControlLabel>Download Data &nbsp;</ControlLabel>
```
rod-glover: Again, it's preferable to use styling to determine layout. The `&nbsp;` isn't really content, it's a forcing for layout.

src/core/export.js
```diff
@@ -0,0 +1,255 @@
...
+  // create workbook object containing one or more worksheets
+  var wb = {
+      Sheets: {},
+      SheetNames: []
+  };
+
+  // prepare filename, metadata cells, and data cells according to type of export
+  var summaryCells, dataCells, outputFilename;
+  var filenamePrefix = "PCIC_CE_";
+  var filenameInfix = "Export_" + metadata.model_id + "_" + metadata.experiment + "_" + metadata.variable_id;
```
rod-glover: Prefer [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) to manual concatenation of strings. This would look like:

```
`Export_${metadata.model_id}_${metadata.experiment}_${ metadata.variable_id}`
```

rod-glover: Oh, I see you use template literals elsewhere ... good.

corviday: Learned of them from one of your earlier reviews. :)

rod-glover:
Nice discussion of tradeoffs of download data options.

rod-glover:
Oops.

rod-glover:
LGTM (looks good to me -- an acronym we used a lot in my previous workplace)

jameshiebert:
> data for export is now parsed from the C3 graph configuration objects

I'm open to discussion on this, but this seems like a non-ideal choice to me. The graph is supposed to be only a consumer of information and getting the table data out of the graph kind of muddies the data flow a bit. It also creates unwanted dependencies on C3 *and* the graph. For example, what if we wanted to swap out C3 for another plotting library? If the table is dependent on C3's data representation, now we have to rewrite the table code too. What if we want a page *without* a graph (and only a table). That's a feasible scenario as well.

I can see the advantage of minimizing API calls, but OTOH, we could more gracefully handle this (or let the *browser* handle this) with a pretty minimal cache-control policy.

Again, I'm open to discussion, so would you mind commenting on how you see this playing out?

Everything else looks great! :)

corviday:
> The graph is supposed to be only a consumer of information and getting the table data out of the graph kind of muddies the data flow a bit.

The graphs are passive consumers of information. 

DataController makes calls to the API and parses the result into objects which it saves as attributes of its own state (`state.climoSeriesData`, `state.timeSeriesData`); this is the only way the API data is saved in DataController's state at present. Those objects are passed by DataController to the DataGraphs as props, and also passed by DataController to the data export functions as arguments.

Maybe that's a distinction without a difference, though.

> If the table is dependent on C3's data representation, now we have to rewrite the table code too. What if we want a page without a graph (and only a table). That's a feasible scenario as well.

The displayed DataTable is not dependant on C3's data representation. 

`DataController` makes calls to the API and parses the result into an object which it then saves as an attribute of its own state (`state.statsData`). This object is passed by DataController to the DataTable as props, and also by DataController to the data export function as an argument. 

(Unlike the graph objects, it has little to no formatting-related content. The table doesn't have a lot of formatting options the way a graph does. So this is arguably less weird than passing the graph props object to the export function.)

(I didn't describe the table's export dataflow explicitly in my "what's changed" post because I meant it to be "what's changed since commit 9134397." I can see how that would have been misleading, given how much everything got refactored and rearranged, sorry.)

> For example, what if we wanted to swap out C3 for another plotting library?

We would have to rewrite/replace `util.parseTimeSeriesForC3()`, `util.dataApiToC3()`, and `export.generateDataCellsFromC3Graph()`, and their various helper functions. Changes to `DataController` and `export.exportDataToWorksheet()` would be fairly minor, I think, mostly just switching in the names of the newly written parser functions.

I've been thinking of this feature as "download the data you see in the graphs" so it doesn't seem too onerous or weird for me that if we have different graphs we'd have to rewrite the code to download from them, but perhaps I'm approaching this with the wrong mindset.

> I can see the advantage of minimizing API calls, but OTOH, we could more gracefully handle this (or let the browser handle this) with a pretty minimal cache-control policy.

Possible options going forward:
1. Leave it as is
2. Make an API call for each download, have a separate helper function to parse the API call into tabular form. (as of commit 9134397ish, with some of the improvements made since then)
3. Have DataController parse data API calls it makes into two separate objects to store in its state, one object to pass to graphs as props and a second simpler object (perhaps just a 2D array) to pass to export functions if a user downloads the data
4. Have DataController parse the graph props object and pass the export function only the bits it needs and not things like, I dunno, tooltip formatting information. I think this is a solution for a problem we're not having, though.
5. I don't know much about how caching would work, but I think it would mean DataController was able to retrieve/remember the API results? So it would be like the 2, but faster and more elegant.

jameshiebert:
> The graphs are passive consumers of information.

OK, I think that when I read this:

> data for export is now parsed from the C3 graph configuration objects

I misunderstood your initial description of how things worked. If it's the DataController that's retaining the information, then yes, that is exactly what we want.

Carry on :)

PR #61, opened 2017-07-07
Resolves #60

### LINKED ISSUE #60: Intermittent failure of TravisCI tests
Occasionally, tests run by TravisCI against a push or pull request made to this repository fail with a syntax error. Jasmine runs tests in an arbitrary order; when a test fails, it is always the first one, and all subsequent tests pass. 

Making a commit with trivial changes, such as adding spaces to a string literal or comment, sometimes results in a previously-failed test passing.

corviday:
Requires jest to run tests serially instead of parallelizing them. Hopefully resolves the ongoing issue with intermittent CI test failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "webpack-dev-server --host 0.0.0.0",
     "build": "webpack --progress --colors",
-    "test": "jest --verbose",
+    "test": "jest --verbose --runInBand",
     "lint": "eslint .",
     "lint:glob": "eslint"
   },


jameshiebert:
This seems like a reasonable approach. We're not doing hardcore UI testing (yet), that requires a ton of processor time, so we won't lose much by running in serial. We can revisit this issue if speed becomes a problem, but for now the testing that you have done is pretty good evidence that this will workaround the root problem (a race condition of some kind). If it continues to be a problem, we can investigate whether the root cause is getting fixed in a more recent version of Jest. In any case, great work!

PR #62, opened 2017-07-22
Isolines and data files separated by time resolution

corviday:
**EDIT: this branch is now ready for a merge-oriented review.**

This isn't ready to merge yet, but I'd appreciate some feedback [on `chart.js` specifically](https://github.com/pacificclimate/climate-explorer-frontend/compare/feature/56?expand=1#diff-f5690561d8dae3aaac74253cbc807376) at this juncture. Chart.js contains `timeSeriesToAnnualCycleGraph()` and its helper functions, which create an annual cycle graph from multiple timeseries. The impetus for this rewrite was the fact that monthly, yearly, and seasonal data are now separate series, but it should produce a reasonable graph from arbitrary timeseries combinations. (For example, comparing two variables on the isoline view.)

I'm particularly interested in readability/clarity, since between the arbitrary C3 object format requirements and the arbitrary 17-point chronology requirements, it's been tough to write clear graphing code on this project. I've tried to be clear about what's going on and why, and would like to hear about anywhere it isn't.


src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+    throw new Error("Error: cannot graph identical timeseries");
+  }
+  
+  //Remove unique_id from the list of possible variations. All
+  //datasets have unique unique_id's; it's not useful on a graph
+  variation.splice(variation.indexOf("unique_id"), 1);
+  
+  //Remove variable_name; it's redundant with variable_id (and takes
+  //up a lot of room.)
+  variation.splice(variation.indexOf("variable_name"), 1);
```
corviday: There needs to be a check that `variable_name` is actually in `variation` before attempting to remove it. Unlike `unique_id`, `variable_name` is not always present.

jameshiebert:
I'll go through this in a bit more detail, but one honest question is: how much of this is a "rewrite" vs. "new functionality". I think that you've billed it as a rewrite, but I'm suspicious of adding 1300 LOCs without the balance of some code removal. Are there things where you have completely replaced their use and could be removed? Or is there now psudo-duplicated functionality that may confuse future programmers on this code base? Again, I'm asking out of ignorance having not delved into the codebase (yet), so it's an honest question.

corviday:
Good question! It's intended to replace `util.parseTimeSeriesForC3()` and its helper functions, as well as `util.mergeC3DataGraphs()`. I haven't deleted them yet, because I wanted to get feedback that approach followed in the new version _is_ actually an improvement. :)

(There are also some functional differences; `util.parseTimeSeriesForC3()` expects a 17-point chronology. `mergeC3DataGraphs()` is hopefully made redundant by just generating a graph with two datasets in the first place, not generating two seperate graphs and merging them.)

corviday:
Oh, wait, you mean the whole branch, not just chart.js? Yeah, there's a lot of new functionality in the whole branch - isoline stuff along with the changes needed to deal with split datasets necessitated by isolines. Most of it isn't ready for review, though. I just wanted a couple extra pairs of eyes on the graph rewrites in `chart.js`.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+ *   "2085-02-15T00:00:00Z": -15.54878007851129,
+ *   "2085-03-15T00:00:00Z": -11.671093808333737,
+ *                    ...
+ *    }
+ * }
+ * 
+ * and return a C3 graph object displaying them.
+ * 
+ * It takes an arbitrary number of data objects, but no more than
+ * two separate unit types. Allowable data resolutions are monthly(12),
```
jameshiebert: Does it allow both timeseries and climatologies? By the 12/4/1, I would assume only climatologies, but it would be good to clarify.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+    var timeseries = data[i];
+    var timeseriesMetadata = _.find(metadata, function(m) {return m.unique_id === timeseries.id;});  
+    var timeseriesName = getTimeseriesName(timeseriesMetadata);
+       
+    //add the actual data to the graph
+    c3Data.columns.push([timeseriesName].concat(getMonthlyData(timeseries.data, timeseriesMetadata.timescale)));
+    
+    //monthly data is displayed as a line graph, but yearly and seasonal
+    //display as step graphs.
+    c3Data.types[timeseriesName] = timeseriesMetadata.timescale == "monthly" ? "line" : "step";
```
jameshiebert: Yep, that works.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+
+    //Each timeseries needs to be associated with a y-axis.
+    //Two different variables measured with the same units, 
+    //like tasmin (degrees C) and tasmax (degrees C)
+    //can share a y-axis. 
+    //Variables with different units require seperate axes.
+    //C3 can theoretically support indefinite numbers of y-axes,
+    //but that would be hard for a user to make sense of, 
+    //so it's capped at two here.
```
jameshiebert: Good idea.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+    if((!yUnits) || timeseries.units == yUnits) {
+      yUnits = timeseries.units;
+      c3Data.axes[timeseriesName] = "y";
+    }
+    else if((!y2Units) || timeseries.units == y2Units) {
+      y2Units = timeseries.units;
+      c3Data.axes[timeseriesName] = "y2";
+    }
+    else {
+      throw new Error("Error: too many data axes required for graph");
```
jameshiebert: Reasonable to throw and error here and not assume responsibility. Are you making sure that we catch it somewhere at a higher level?

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+  return {
+    data: c3Data,
+    tooltip: c3Tooltip,
+    axis: c3Axis
+  }; 
+};
+
+/*
+ * Helper function for parseTimeseriesForC3.
+ * Accepts a dataseries object with 1, 4, or 12 timestamp:value pairs
```
jameshiebert: Hmmm, so the idea is that monthly is the lowest common denominator for all of the supported data types (monthly/seasonal/annual climatologies) and you'll downscale everything that's coarser to monthly? That seems like an OK strategy for 1/4/12/17 resolution, but I definitely would not want to do this for higher frequency timeseries (e.g. a 150 year annual timeseries). Keep those use cases in mind for the future.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+ * need to appear in the graph legend. But if graphing multiple variables,
+ * the graph legend will need to display variable names as well.
+ * 
+ * Timeseries names include any descriptive  metadata that vary between 
+ * timeseries and leave out any metadata that doesn't. They end with "mean".
+ */
+//TODO: special case climatological period to display as (XXXX-XXXX)
+//TODO: possibly cue descriptors to appear in a specific order?
+// "Tasmin Monthly Mean" sounds better than "Monthly Tasmin Mean".
+var shortestUniqueTimeseriesNamingFunction = function (metadata, data) {
```
jameshiebert: This is an interesting function, and I think that it's a really nice way to separate out some funky heuristics that will *really* improve usability and readability of the charts. It *does* have a lot of weird cases though, so if you don't have tests in place yet, I'd advise that you write some.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+    name = name.concat("mean");
+    return capitalizeWords(name);
+  };
+}
+
+/* 
+ * Simple formatting function for numbers intended for display.
+ */
+//TODO: look up significance algorithms and write a smarter one!
```
jameshiebert: This is how we do it in ``pacificclimate/pdp``: https://github.com/pacificclimate/pdp/blob/master/pdp/static/js/pdp_controls.js#L215

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+ * }
+ * 
+ * and return a C3 graph object displaying them.
+ * 
+ * It takes an arbitrary number of data objects, but no more than
+ * two separate unit types. Allowable data resolutions are monthly(12), 
+ * seasonal (4), or yearly (1); an error will be thrown 
+ * if this function is called on data with another time resolution.
+ */
+var timeseriesToAnnualCycleGraph = function(metadata, ...data) {
```
rod-glover: Just a little aside: Is there some reason we prefer this `var f = function(args) ...` style declaration to the simpler `function f(args) ...` style?

src/core/chart.js
rod-glover: My thought exactly. Please add a comment to that effect here.

src/core/chart.js
rod-glover: Agree. JavaScript is very good at sailing invisibly over errors, which, uncaught, appear only in the console.

rod-glover: Also, nitpick: No need to repeat the word "Error" in the string in the Error object.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+  var expectedTimestamps = {"monthly": 12, "seasonal": 4, "yearly": 1};
+  var monthlyData = [];
+  var timestamps = Object.keys(data).sort();
+  
+  if(timestamps.length == 17) {
+    throw new Error("Error: concatenated 17-point chronology.");
+  }
+  
+  if(timestamps.length != expectedTimestamps[timescale]) {
+    throw new Error("Error: inconsistent time resolution in data");
```
rod-glover: Also, what catches these errors and what is done about it? Sorry if the answer ought to be obvious.

src/core/chart.js
```diff
@@ -0,0 +1,252 @@
...
+  }
+  
+  for(var i = 0; i < 12; i++) {
+    var mapped = Math.ceil((timestamps.length / 12.0) * (i + 1)) - 1;
+    monthlyData.push(data[timestamps[mapped]]);
+  }
+  
+  //Seasonal timeseries need one month of winter removed from the beginning of the
+  //year and added at the end, since winter wraps around the calendar new year.
```
rod-glover: :+1:

src/core/chart.js
corviday: Errors thrown here are caught by `DataController` and displayed to the user via `DataController.displayError()`. They appear on the graph instead of data. That is, the graph displays "Loading" while the API is queried and the graph is generated, and then either data or an error message is displayed.

I like having "Error:" in the error messages that will be displayed on the graph in place of data because I feel it makes it clearer to the user that their data isn't going to show up eventually if they wait, but I'm not too attached to the idea.

src/core/chart.js
corviday: Thanks! That's really helpful.

src/core/chart.js
corviday: `DataController` and its siblings (`DualDataController` and `MotiDataController`) catch these errors. The only thing that's done with them is that their text is displayed to the user, on the graph, in place of the expected data visualization.

corviday:
This branch still isn't completely ready for merging, but it's getting close. The functionality is pretty much there, but some cleanup and testing are needed.

Outstanding task list:
**Functionality**
- [ ] Find intermittent isoline map data selector bug (or verify it's an ncWMS jar issue, if it is)
- [ ] Possibly make map timestamp selection prettier?
- [ ] Use climatology period from metadata query until multimeta provides it

**Cleanup**
- [ ] make dataApiToC3() more flexible, fewer built-in assumptions
- [ ] Change every variable named `*TimeSeries*` to `*Timeseries*`
- [ ] Look at util.js and remove any functions that are no longer being used
- [ ] Follow JSHint's suggestions
- [ ] Organize function order in util.js and export.js

**Tests**
- [ ] Fix the warning message on the Selector tests
- [ ] Write tests for new functions in util.js or chart.js that could use one
- [ ] Fix broken bootstrap table formatting test

corviday:
This branch is ready for review with an eye to merging.

This is a fairly extensive set of changes that, unfortunately, couldn't be made in separate branches without breaking things. They fall into four themes and can probably mostly be considered separately for review purposes:

**1. New Components To Display Isolines**
 - [DualController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-100a1154fd1f260b9facd5547fd879c9)
- [DualController.css](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-368f81cb498cc89eb00bffa1c43a7487)
 - [DualController/package.json](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-8753a43a2260c1f0b479222cba166ea6)
 - [DualDataController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-a44da5792da1a6cad95fb54bd5e30216)
 - [DualDataController.css](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-838332bc789341678ba99954c7b19f0b)
- [DualMapController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-923d7065ce1769f41afea095072d53c2)
 - [DualMapController.css](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-0a616735c96c6b9726f5e9e45a0b9cdc)
 - [DualMapController/package.json](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-af032625724ca5f9c41edfd34eb778de)
 - [DualMap.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-182b3090f5e43aeb0030d2b3c7ac5e5a)
 - [index.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-1fdf421c05c1140f6d71444ea2b27638)



**2. Existing Components Adjusted To Handle Data Files Separated By Time Resolution**
 - [DataController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-f4fb9317dcd3771e665609064ef1de92)
 - [DataControllerMixin.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-228f81452c16317a214e8dba7e0b5c09)
 - [MotiController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-e5fcfb3b2179c7ab569be89444ab1396) - actually, all I seem to have done here is accidentally add whitespace, will fix.
 - [MotiDataController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-c7ffa98c0eab963cdf2f0c61cff90a56)
 - [TimeOfYearSelector.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-9957e4ff97331ddc26a412c6edf65d3a)
 - [util.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-7c30ff4316a8c7ca19fe452f77768c8d) - (parts spun off into chart.js)
 - [chart.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-f5690561d8dae3aaac74253cbc807376) (split out from util.js, which was getting unwieldy)
- [export.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-304b9ae38198ab8decfc3f64d2312eb4)


**3. Existing Components Adjusted To Handle New Version Of ncWMS**
 - [CanadaMap.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-92d4b6a4d6f6eb6c3055a5bbe10e4bd8)
 - [MapController.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-33b1171fbafdabe1d65f669da0852488)
 - [leaflet-ncwms-autoset-colorscale.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-304b9ae38198ab8decfc3f64d2312eb4)
 - [leaflet-ncwms-colorbar.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-92c9c00d8918f2befaf859ad33861f47)


**4. Tests and Cleanup To Go With 1-3**
 - [AppMixin.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-87e470a26978ea8302cfdd3d2f8f31ea)
- [chart-test.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-0066a17f5171f1f55a192464a08c9828)
- [export-data.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-0066a17f5171f1f55a192464a08c9828) - old sample data for testing, outdated now that our data structure has changed
 - [export-test.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-70b9bdbf7e8f4ce84e7bac244b5a515a)
 - [sample-API-results.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-2229c2f2cc9a400ac3d87acc7ccbdc06) - sample data for test that matches new data structure
- [test-validators.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-eaed2c41599b28c348ddb9b97b5b6885)
 - [util-tests.js](https://github.com/pacificclimate/climate-explorer-frontend/pull/62/files#diff-09a3c7bd529ce40c006291faabea5463)

rod-glover:
There does seem to be a lot of overlap between the "dual" controllers and the single controllers. As we discussed, it may now be worth coalescing these into a single controller with selectable behaviour. This poses risks if the requirements of the two display types they support diverge in future. OTOH, maintaining copies of code and fixing bugs in multiple places is a risk too.

rod-glover:
Your call on this one. You've already done a lot of great work here, and it may be better to coalesce in a separate branch after this one is merged.

corviday:
I think there's a stronger case for merging the two `MapControllers` than the three `AppControllers` or `DataControllers`.

1. The `DataControllers` and the `AppControllers` have mixins, and a lot of their overlapping functionality is already there. The code that differs is mostly UI-related, the logic and data-manipulation are shared. The `MapControllers` don't have a mixin or any other measure taken to reduce repetitions like that, and are more repetitive.

2. Aside from that, I think the two `MapControllers` might have fewer differences. I expected them to diverge more widely when I first starting working on this project, I think.

3. Also, multiple `MapControllers` is a new phenomenon and didn't exist before this branch. So it's a question of "continue living with already-existing questionable design decisions" vs "make a brand new questionable design decision". :)

corviday:
I guess I've talked myself into at least looking at how hard it would be to reduce the number of MapControllers.

src/core/__tests__/export-test.js
```diff
@@ -0,0 +1,142 @@
...
+var mockAPI = require('./sample-API-results');
+var util = require('../util');
+var chart = require('../chart');
+
+/*
+ * It's not clear how to test exportDataToWorksheet, which would require
+ * capturing and testing a file saved to the user. However, most of the 
+ * work of creating the files is done by helper functions that can be 
+ * tested seperately, so overall there's reasonable test coverage of 
+ * export functionality.
```
rod-glover: :+1:

src/core/__tests__/export-test.js
```diff
@@ -0,0 +1,142 @@
...
+ */
+describe('exportDataToWorksheet', function () {
+  xit('exports an annual cycle graph to file', function () {});
+  xit('exports a stats table to file', function () {});
+  xit('exports a projected change graph to file', function () {});
+});
```
rod-glover: cute

rod-glover:
I've reviewed as much as I feasibly can today. Most of my comments are pretty minor, excepting questions about coalescing common code. 

A lot of good work here. I particularly appreciate the more extensive comments and the breaking out of items into utility packages instead of packing a component with things that aren't really its direct responsibility.

corviday:
At present, this branch requires [the multimeta API call to return period information](https://github.com/pacificclimate/climate-explorer-backend/pull/62), which is not yet available in the backend's master branch.

corviday:
Bugs turned up during demo for Trevor:

- [ ] dataset selection UI element doesn't update properly
- [ ] switching to linear or logarithmic values on the map temporarily loses a layer

corviday:
Working on the logscale bug discovered during the demo this morning. Here are two calls to ncWMS:

http://docker2.pcic.uvic.ca:8600/ncWMS/wms?SERVICE=WMS&REQUEST=GetMap&VERSION=1.1.1&LAYERS=tasmin_aClim_BCCAQv2_bcc-csm1-1-m_historical-rcp45_r1i1p1_19610101-19901231_Canada%2Ftasmin&STYLES=default-scalar%2Fseq-Greys&FORMAT=image%2Fpng&TRANSPARENT=true&HEIGHT=256&WIDTH=256&NOWRAP=true&TIME=1977-07-02T00%3A00%3A00Z&NUMCOLORBANDS=249&SRS=EPSG%3A4326&LOGSCALE=false&BBOX=-125.00000000000001,65,-100,90

http://docker2.pcic.uvic.ca:8600/ncWMS/wms?SERVICE=WMS&REQUEST=GetMap&VERSION=1.1.1&LAYERS=tasmin_aClim_BCCAQv2_bcc-csm1-1-m_historical-rcp45_r1i1p1_19610101-19901231_Canada%2Ftasmin&STYLES=default-scalar%2Fseq-Greys&FORMAT=image%2Fpng&TRANSPARENT=true&HEIGHT=256&WIDTH=256&NOWRAP=true&TIME=1977-07-02T00%3A00%3A00Z&NUMCOLORBANDS=249&SRS=EPSG%3A4326&LOGSCALE=true&BBOX=-125.00000000000001,65,-100,90

They are completely identical except the` LOGSCALE` argument, but one of them returns the expected image, the other an XML parsing error. Odd.

corviday:
Aha! Figured it out. You can't use logscale on any data that contains negative values. So during the demo, we were only apply to apply it to precipitation, which cannot be negative.

jameshiebert:
From a programming perspective, I like something between 2 and 3. 2 won't necessarily work 100% of the time with climate data because a) log scaled won't work with data == 0 (which precip often is) and b) some models may show negative precip I think.

I'd like something such as: default to using minmax but whitelist a few variable names to enable log? Keep in mind that you can set WMS params to limit the range of the plot (I.e. Set a minimum to trace for log requests).

corviday:
Huh, neat. What would negative precip mean in the context of a model that outputs it?

corviday:
Since this branch is already 3k-ish lines, I've decided logscale functionality and associated complexity belongs on another branch. I took the logscale UI elements off the MapController dialogue and incorporated Rod's excellent suggestions.

src/components/MapController/MapController.js
```diff
@@ -19,51 +38,98 @@ var MapController = React.createClass({
...
+  
+  //this function handles user selection of all the straightforward
+  //parameters like logscale or palette.
   updateSelection: function (param, selection) {
     var update = {}; update[param] = selection;
     this.setState(update);
   },
 
+  //this function updates the timestamp displayed on the map and triggers 
+  //a rerender.
```
rod-glover: Technically, this code only updates the time values in the state. The consequences of that are to update the map display and trigger a renderer. In fact, the consequences of that update could change (if other code changes), presumably, so better to state this as 'nominally ...' or 'typically ...' or omit entirely unless this consequence is critical to understanding something here.

src/components/MapController/MapController.js
```diff
@@ -19,51 +38,98 @@ var MapController = React.createClass({
...
-    // Updates dataset in state. Updates time value to match new dataset
-
-    this.selectedDataset = this.props.meta.filter(function (el) {
-      return el.unique_id === uniqueId;
-    })[0];
-
-    this.requestTimeMetadata(uniqueId).then(response => {
-      this.selectedDataset.times = response.data[uniqueId].times;
+  //this function stores a dataset selected by the user to state, 
+  //triggering a rerender to show the new dataset.
```
rod-glover: Similar comment to just above. It's a bit clearer here, but documenting the likely effects of an update in a pub-sub system is perilous, and establishes, in comments, a coupling between parts that pub-sub by design decouples.

src/components/MapController/MapController.js
```diff
@@ -19,51 +38,98 @@ var MapController = React.createClass({
...
+          var idxString = JSON.stringify({
+            "timescale": responses[i].data[id].timescale,
+            "timeidx": timeidx,
+          });
+          times[idxString] = responses[i].data[id].times[timeidx];
+        }
+      }
+      //select an arbitrary index to display initially. It will be a
+      //0th index, but could be January, winter, or Annual.
+      var startingIndex = Object.keys(times)[0];
```
rod-glover: Danger! A hash is unordered. Hash keys are not guaranteed to be returned in the order they are stored. So the value initially displayed could be any of the values stored. 
If that's OK behaviour, this code is OK. If not (and it seems a bit weird to allow an effectively random initial value), you need to add some kind of order-awareness -- perhaps the simplest would be to store the first index as it is generated.

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
+      cPalette = nextProps.contourPalette;
+    }
+    else if(this.state.scalarPalette) {
+      sPalette = this.state.scalarPalette;
+      cPalette = this.state.contourPalette;
+    }
+    else {
+      sPalette = nextProps.comparandMeta ? 'seq-Greys' : 'x-Occam';
+      cPalette = nextProps.comparandMeta ? 'x-Occam': undefined;
+    }
```
rod-glover: Hmm, a trifle tricky to understand. I'd go with a single if-else statement here instead of the two conditional assignments.

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
+   
+    //generate a list of timestamps available for the specified data
```
rod-glover: From here ...

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
+            "timescale": responses[i].data[id].timescale,
+            "timeidx": timeidx,
+          });
+          times[idxString] = responses[i].data[id].times[timeidx];
+        }
+      }
+
+      //select a starting timestamp to display.
+      //should correspond to a 0th index, but might be annual, seasonal, or monthly
+      var startingTime = Object.keys(times)[0];
```
rod-glover: ... to here is AFAICT identical to most of `updateDataset` above. 

Factor that out into a shared function? If you don't, please apply the comment there about unordered hash keys here too.

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
-    //FIXME: Time period should be determined from the metadata API
-    // which currently doesn't give time bounds information. See here:
-    // https://github.com/pacificclimate/climate-explorer-backend/issues/44
-    // When that issue is fixed, this code needs to be updated
+    //run and period selector
+    //displays a list of all the unique combinations of run + climatological period
+    //a user could decide to view.
+    //Not sure JSON is the right way to do this, though.
+    //TODO: see if there's a more elegant way to handle the callback 
+    //(selector won't pass an object)
```
rod-glover: Agree could be more elegant.

The obvious alternative (but not thought through here in detail) would be to use an index into a list (whether implicit -- you'd need to establish a canonical ordering for that --, or explicit -- you'd need to share a data structure for that).

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
+      end_date: this.state.end_date,
+      ensemble_member: this.state.run
+    });
     if (ids.length > 1) {
       datasetSelector = (<Selector
         label={"Select Dataset"}
         onChange={this.updateDataset}
-        items={ids} value={this.state.dataset}
+        items={ids} value={`${this.state.run} ${this.state.start_date}-${this.state.end_date}`}
+        value={selectedDataset}
```
rod-glover: Why is `value` specified twice here? And what does that mean?

corviday: Whoops, I think it means I wasn't paying enough attention. :)

src/components/MapController/MapController.js
```diff
@@ -89,93 +157,220 @@ var MapController = React.createClass({
...
+    if(this.props.comparandMeta) {
+      contourPaletteSelector = (            
+        <Selector
+          label={"Isoline Colour Palette"}
+          onChange={this.updateSelection.bind(this, 'contourPalette')}
+          items={palettes}
+          value={this.state.contourPalette}
+        />
+      );
+      //TODO:
```
rod-glover: TODO?

src/components/Map/CanadaMap.js
```diff
@@ -221,29 +262,68 @@ var CanadaMap = React.createClass({
...
+      }));
+    }
+    else if(this.props.contourDataset) {
+      map.addControl(new NcWMSAutoscaleControl(this.ncwmsContourLayer, {
+        position: 'bottomright'
+      }));
+      map.addControl(new NcWMSColorbarControl(this.ncwmsContourLayer, {
+        position: 'bottomright'
+      }));
+    }
```
rod-glover: I _think_ the following might be easier to read and less repetitive:

```
    if(this.props.scalarDataset) {
      scalarAs = new NcWMSAutoscaleControl(this.ncwmsScalarLayer, {
        position: 'bottomright'
      });
      scalarCb = new NcWMSColorbarControl(this.ncwmsScalarLayer, {
        position: 'bottomright'
      });
    }
    
    if(this.props.contourDataset) {
      contourAs = new NcWMSAutoscaleControl(this.ncwmsContourLayer, {
        position: 'bottomright'
      });
      contourCb = new NcWMSColorbarControl(this.ncwmsContourLayer, {
        position: 'bottomright'
      });
    }
    
    if(this.props.contourDataset && this.props.scalarDataset) {
      map.addControl(scalarCb);
      scalarAs.addLayer(this.ncwmsContourLayer);
      map.addControl(scalarAs);
      map.addControl(contourCb);  
    }
    else if(this.props.contourDataset) {
      map.addControl(contourAs);
      map.addControl(contourCb);
    }
    else if(this.props.scalarDataset) {
        map.addControl(scalarAs);
        map.addControl(scalarCb);
    }
```

What do you think?

rod-glover: Or better yet:

```
    function makeControls(layer) {
      return [
        new NcWMSAutoscaleControl(layer, {
            position: 'bottomright'
        }),
        new NcWMSColorbarControl(layer, {
            position: 'bottomright'
        })
      ]

    }

    if(this.props.scalarDataset) {
      [scalarAs, scalarCb] = makeControls(this.ncwmsScalarLayer);
    }

    if(this.props.contourDataset) {
      [contourAs, contourCb] = makeControls(this.ncwmsContourLayer);
    }

    ...
```

rod-glover:
Reviewed MapController.js and CanadaMap.js. A few suggestions, otherwise looking very good.

src/components/Map/CanadaMap.js
```diff
@@ -1,3 +1,26 @@
+/***************************************************************
+ * CanadaMap.js - display component using ncWMS and Leaflet.
+ * Will display up to two map layers, one as shaded color blocks, 
+ * and one as isolines, or either independently, over a tiled 
+ * background. Origin and bounds display all of Canada.
+ * 
+ * Child of (and viewer for) MapController.
+ * 
+ * Props passed to configure shaded color map:
```
jameshiebert: This isn't a comment on the code necessarily, but why are you naming all of the raster layer with the prefix "scalar". "Scalar" to me implies one single number, and in my mind (and the language we use), we've never called that map a scalar map...

src/core/leaflet-ncwms-autoset-colorscale.js
```diff
@@ -38,21 +44,31 @@ var ncWMSAutoscaleControl = L.Control.extend({
...
-        bbox: this.layer._map.getBounds().toBBoxString(),
-        time: this.layer.wmsParams.time,
-        srs: this.layer.wmsParams.srs,
-        width: 100,
-        height: 100,
-      },
-    }).then(response => {
-      this.layer.setParams({ colorscalerange: response.data.min + ',' + response.data.max });
-    });
+    for(var i = 0; i < this.layers.length; i++) {
```
jameshiebert: Kind of a nitpick, but why can't you do ``this.layers.forEach()`` here as well?

jameshiebert:
@rod-glover You had your requested suggestions addressed, correct?

rod-glover:
@corviday can we review these changes side-by-side? I'm having trouble determining what changes you made in response to my comments. This is by no means a criticism, it's just that this is a huge and long-running PR and it's hard to figure out.

rod-glover:
In general I think this is a very good piece of work.

src/components/Map/CanadaMap.js
corviday: ncWMS2 calls the shaded colour map style "scalar," so I was using "scalar" to distinguish that ncWMS raster layer from the other "contour" styled (isoline) ncWMS raster layer. I'm not sure why ncWMS calls that layer style "scalar".

Suggestions for more intuitive-to-us names for the two ncWMS layers? Colours & isolines? Raster & isolines? Chloropeth and isolines?

jameshiebert: Huh, well if ncWMS calls it that, then I guess that's more sensible. I'd generally vote for raster and isolines, but if you have already used the language of ncWMS, then perhaps we should align with it.

src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -86,62 +109,84 @@ var ModalMixin = {
...
         displayMethod("Error: " + error.response.status + " received from data server.");
-      }else if(error.request) { // data server didn't respond
+      }else if(error.request) { // axios error: data server didn't respond
         displayMethod("Error: no response received from data server.");
-      }else {  // either a failed data validation
-        // or a generic and somewhat unhelpful "Network Error" from axios
+      }else {  
+        // either an error thrown by a data validation function,
+        // an error thrown by the DataGraph or DataTable parsers,
+        // or the generic and somewhat unhelpful "Network Error" from axios
```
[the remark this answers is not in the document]
corviday: Usually axios provides an HTTP status code in `error.response`. Once in a while it doesn't, and provides only the "Network Error" message without populating `error.response` or `error.request`. [We are not the only ones who have encountered intermittent lack of http status on axios errors.](https://github.com/mzabriskie/axios/issues/383)

As far as I could tell from testing, the lack-of-http-status-code in our case happened when the backend's status was 500, but not every time. Perhaps it was related to how long the backend took to return a response.

This may be fixed, I see axios' most recent update [changelog](https://github.com/mzabriskie/axios/blob/master/CHANGELOG.md) (0.16.2) includes
- Including underlying request in errors

PR #65, opened 2017-09-20
Error message when no data available. Fix #63

### LINKED ISSUE #63: Gracefully handle nonexistent model / experiment / variable combinations.
The model, experiment, and variable selectors at the top of the /climo and /compare pages contain all models, experiments, and variables from the database, but it's possible to select a combination that describes no datasets. 

Example: if all CanESM2 run in the metadata database are based on  hist+rcp45, but all  bcc-csm2-m-1 runs are based on hist+rcp26, a user could select CanESM2 as their model and hist+rcp26 as their experiment, filtering out all datasets.

Right now, this results in nothing happening - previously displayed datasets from a valid model/experiment/variable configuration continued to be displayed and can be interacted with, which seems like it would be confusing (though there is a title about the graphs that will continue displaying the parameters of the old, valid dataset).

corviday:
Clears maps and graphs when user selects a model/variables/emission scenario combination that is not present in the database. Before this update, the previously selected valid data would continue to be displayed, possibly causing confusion.

[Test docker for this PR](http://docker2.pcic.uvic.ca:8604/climo). Changing either the model or the emissions scenario away from starting values demonstrates the new behaviour.

### DIFF
--- a/src/components/DataControllerMixin/DataControllerMixin.js
+++ b/src/components/DataControllerMixin/DataControllerMixin.js
@@ -38,11 +38,26 @@ var ModalMixin = {
 
 
   componentWillReceiveProps: function (nextProps) {
-    this.setState({
-      timeSeriesDatasetId: nextProps.meta[0].unique_id,
-    });
-    if (this.verifyParams(nextProps)) {
+    if (this.verifyParams(nextProps) && nextProps.meta.length > 0) {
+      this.setState({
+        timeSeriesDatasetId: nextProps.meta[0].unique_id,
+      });
       this.getData(nextProps);
+    }
+    else { //Didn't receive any valid data.
+      //Most likely cause in production would be the user selecting
+      //parameters (rcp, model, variable) for which no datasets have been
+      //added to the database yet.
+      //In development, could be API or ensemble misconfiguration, database down.
+      //Display an error message on each viewer in use by this datacontroller.
+      var text = "No data matching selected parameters available";
+      var viewers = ["StatsTable", "ClimoSeries", "TimeSeries"];
+      _.each(viewers, function(viewer) {
+        var viewerFunction = `set${viewer}NoDataMessage`;
+        if(typeof this[viewerFunction] == 'function'){
+          this[viewerFunction](text);
+        }
+      }, this);
     }
   },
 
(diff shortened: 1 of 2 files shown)


src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -38,12 +38,27 @@ var ModalMixin = {
...
+      //added to the database yet.
+      //In development, could be API or ensemble misconfiguration, database down.
+      //Display an error message on each viewer in use by this datacontroller.
+      var text = "No data matching selected parameters available";
+      var viewers = ["StatsTable", "ClimoSeries", "TimeSeries"];
+      _.each(viewers, function(viewer) {
+        var viewerFunction = `set${viewer}NoDataMessage`;
+        if(typeof this[viewerFunction] == 'function'){
+          this[viewerFunction](text);
+        }
```
rod-glover: Wow, that is some JS kung fu. Although it's repetitive, and clearly that's the point of your code, I might consider just indexing a hash object mapping viewer name to function directly, thus:

```
viewerFunction = {
   'name1': this.setname1NoDataMessage,
   ...
}[viewer]
```

corviday: Thanks for your suggestion! I used it as a starting point, but ended up doing something slightly different, trying to minimize repetition. Would appreciate your stylistic feedback if you have a chance. :)

src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -38,12 +38,28 @@ var ModalMixin = {
...
+      //Display an error message on each viewer in use by this datacontroller.
+      var text = "No data matching selected parameters available";
+      var viewerMessageDisplays = [this.setStatsTableNoDataMessage,
+                                   this.setClimoSeriesNoDataMessage,
+                                   this.setTimeSeriesNoDataMessage];
+      _.each(viewerMessageDisplays, function(display) {
+        if(typeof display == 'function') {
+          display(text);
+        }
+      });
```
rod-glover: Definitely the best. Good call.

jameshiebert:
LGTM!

PR #66, opened 2017-09-20
Logarithmic colour scaling for maps. Resolves #64

### LINKED ISSUE #64: Add ability to view maps with a logarithmic colour scale
NcWMS can generate a map shaded with a logarithmic colour scale for maps where all values are greater than zero. This is most valuable for precipitation, which is skewed in a way a logarithmic scale would reflect well, but might be desired for other variables too.

corviday:
Enables logarithmic colour scaling on the map for variables where either:

- All values of the variable are > 0 for the timeslice currently being displayed
- The variable is listed in` variable-options.yaml` with an `overrideLogarithmicScale` value of `true`, in which case, this particular variable _always_ has the option to be clipped to > 0 and displayed with logarithmic colouring

Also added a couple other low-hanging items mentioned by Trevor during the isolines demo to `variable-options.yaml`:
 - specify a default colour scheme to display a variable on the map
- specify a default precision for a variable, which is used in both graphs and maps. 

[Test docker for this branch](http://docker2.pcic.uvic.ca:8605/compare). Precipitation has a default colour scheme and logarithmic override enabled. Tasmin and tasmax have a decimal precision (1) specified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "PCIC Climate Explorer front end application",
   "main": "index.js",
   "scripts": {
+    "prestart": "if [ ! -f ./variable-options.yaml ]; then touch ./variable-options.yaml; fi",
     "start": "webpack-dev-server --host 0.0.0.0",
     "build": "webpack --progress --colors",
     "test": "jest --verbose --runInBand",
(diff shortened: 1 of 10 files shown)


corviday:
Found an issue with logarithmic state "sticking" after switching to a variable that shouldn't allow it.

PR #67, opened 2017-09-20
Add logarithmic scales to map viewer. Fixes #64

### LINKED ISSUE #64: Add ability to view maps with a logarithmic colour scale

corviday:
Enables logarithmic colour scaling on the map for variables where either:

- All values of the variable are > 0 for the timeslice currently being displayed
- The variable is listed in` variable-options.yaml` with an `overrideLogarithmicScale` value of `true`, in which case, this particular variable _always_ has the option to be clipped to > 0 and displayed with logarithmic colouring

Also adds a couple other low-hanging items mentioned by Trevor during the isolines demo to `variable-options.yaml`:
 - specify a default colour scheme to display a variable on the map
- specify a default precision for a variable, which is used in both graphs and maps. 

[Test docker for this branch](http://docker2.pcic.uvic.ca:8605/compare). Precipitation has a default colour scheme and logarithmic override enabled. Tasmin and tasmax have a decimal precision (1) specified.

### DIFF
--- a/src/components/MapController/MapController.js
+++ b/src/components/MapController/MapController.js
@@ -17,7 +17,9 @@
  *******************************************************************************/
 
 import React from 'react';
-import { Row, Col, Button, ButtonGroup, Glyphicon, Modal } from 'react-bootstrap';
+import { Row, Col, Button,
+         ButtonGroup, Glyphicon,
+         Modal, Tooltip, OverlayTrigger } from 'react-bootstrap';
 import urljoin from 'url-join';
 import Loader from 'react-loader';
 import _ from 'underscore';
@@ -263,6 +265,24 @@ var MapController = React.createClass({
     return !_.isEqual(nextState, this.state);
   },
 
+  //This function wraps a React component in a React OverlayTrigger that
+  //displays the supplied text as a tooltip when hovered over.
+  addTooltipWrapper: function (component, text, direction="left") {
+    var tooltip = (
+        <Tooltip id="tooltip">
+          {text}
+        </Tooltip>
+        );
+
+    return (
+        <OverlayTrigger placement={direction} overlay={tooltip} container={this}>
+          <div>
+            {component}
+          </div>
+        </OverlayTrigger>
+    );
+  },
+
   //This function returns JSX for a selector allowing the user to choose
   //whether a map's colours are scaled logarithmically or linearly.
   //If a given map cannot be displayed with logscaled colour, returns an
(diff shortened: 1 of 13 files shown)


jameshiebert:
@corviday Can you resolve the conflicts here?

src/components/MapController/MapController.js
```diff
@@ -225,8 +263,57 @@ var MapController = React.createClass({
     return !_.isEqual(nextState, this.state);
   },
 
+  //This function returns JSX for a selector allowing the user to choose
+  //whether a map's colours are scaled logarithmically or linearly.
+  //If a given map cannot be displayed with logscaled colour, returns an
+  //empty string.
```
jameshiebert: As a user, I found this to be a little confusing. When I went into the settings, I didn't see the selection and was like, "Am I running the right code? Where is the selector?" Would it be better to emit a disabled selector (with the value set to linear) + a tooltip that shows why it is disabled, so that it's more transparent to the user as to why that option is not available?

variable-options.yaml
```diff
@@ -0,0 +1,36 @@
+################################################################
+# variable-options.yaml - variable-specific configuration file
```
jameshiebert: This is a nice way to do it. Good idea.

src/components/MapController/MapController.js
corviday: Good point.

jameshiebert:
Maybe, I'm doing something wrong, but I can't actually get the log scale selectors to be displayed as "disabled". They are either emitted as enabled or not emitted at all. See deployment of docker image [pcic/climate-explorer-frontend:feature_64](https://hub.docker.com/r/pcic/climate-explorer-frontend/builds/brcvx3gmmygot87w7ssor8n/) [here](http://docker2.pcic.uvic.ca:9997/compare).

corviday:
I just built a [docker](http://docker2.pcic.uvic.ca:8601/compare) from this branch, and it seems to be working correctly. Could dockerhub be a commit behind? Is that a thing that can happen?

corviday:
```bash
[lzeman@docker2 ~]$ docker pull pcic/climate-explorer-frontend:feature_64
feature_64: Pulling from pcic/climate-explorer-frontend

Digest: sha256:5dc17c93ba0b3a29aadab0d29179feedb11e0969afae8a5b995ece94737e1b7e
Status: Image is up to date for pcic/climate-explorer-frontend:feature_64

[lzeman@docker2 ~]$ docker run -it -e "CE_BACKEND_URL=http://docker2.pcic.uvic.ca:9999/api" -e "CE_ENSEMBLE_NAME=all_downscale_files" -e "NCWMS_URL=http://docker2.pcic.uvic.ca:9000/ncWMS/wms" -p 8603:8080 --name cef-64-fromhub pcic/climate-explorer-frontend:feature_64

```
Gave me[ this container](http://docker2.pcic.uvic.ca:8603/compare), which also seems to feature disabled logscale dropdowns.

jameshiebert:
Huh. No, that looks good. You win. I ran it with this:

PR #73, opened 2017-09-27
Get export metadata from timeSeriesInstance, fix #71

### LINKED ISSUE #71: Data download of Annual Series graph in /climo portal ignores climatology selected
Data downloaded from the Annual Series graph in the climo portal seems to show the 1961 - 1990 climatology, even if the user has selected a different period. Need to check the `selection` argument in `export.exportDataToWorksheet()`, and see if the problem occurs with any other graphs or portals.

corviday:


src/core/__tests__/export-test.js
```diff
@@ -55,9 +55,10 @@ describe('createTimeSeriesWorksheetSummaryCells', function () {
   it('generates summary cells for an exported annual cycle graph', function () {
     var id = mockAPI.monthlyTasmaxTimeseries.id;
     var metadata = _.find(mockAPI.metadataToArray(), function(m) {return m.unique_id == id;});
+    var run = _.pick(metadata, "start_date", "end_date", "ensemble_member");
```
jameshiebert: The term "run" has a different meaning in climate modelling than it does the way you're using it here. The run (in CMIP3), also known as the realization, is an execution of the model under different (but equally plausible) initial conditions. Probably better to chose another word for this variable to avoid confusion.

corviday: Any suggestions?

PR #75, opened 2017-09-27
Disable webpack-dev-server host checking, fix #69

### LINKED ISSUE #69: Configure webpack to allow host requests inside docker containers

corviday:
As discussed this morning, disable host header checking in the config for webpack-dev-server, as we only use webpack-dev-server in internal development and testing environments that aren't accessible by the outside world.

Tested by spinning up a docker container from this branch, which went uneventfully.


jameshiebert:
Great! Thanks for investigating this!

Issue #74, opened 2017-09-27
Remove serialized attributes passed by dropdown menus


corviday:
In several places in climate explorer, a user selects a combination of two variables from a single dropdown menu, such as start date and end date, or time resolution and time index. The selector then needs to notify a callback function which two (or more) items the user has picked.

As our selectors can only pass atoms to callback functions, it isn't possible to have the dropdown pass something like:
```JSON
{
  "timeres": "monthly",
  "time idx": 4
}
```
Right now dropdowns in climate explorer handle this by serializing the above object and passing it as a string to the callback function to be parsed back into an object, which works adequately, but is not ideal. It would be better to have the dropdowns pass an index to a list containing the relevant attributes instead.

corviday:
Refactored most instances of this antipattern while debugging issues with the TimeOfYearSelector for annual-only datasets (such as climdexes).

There's still a couple instances that pass a serialized object containing a run and a period, but there's a lot less flinging around of serialized objects than there was.

PR #79, opened 2017-09-29
Fix #78

### LINKED ISSUE #78: DualDataController selector does not change
I think that PR #73 may have introduced a bug to the "instance" Selector. The instance shown in the selector is stuck on the first entry.

Steps to reproduce:
1. See the deployment of pcic/climate-explorer-frontend:latest [here](http://docker2.pcic.uvic.ca:9998/compare)
1. Choose the "Annual cycle" plot
1. In the plot selector, select a "Dataset" that is not the first in the list (e.g. 'r1i1pi 2070-2099')

Expected result:
'r1i1pi 2070-2099' will be selected

Actual result:
The first element in the list (in this case 'r1i1p1 1961-1990') is selected

I'm guessing that it's a problem with [line 233 in the render function](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/src/components/DualDataController/DualDataController.js#L233), but I'd be grateful if you could investigate.

FWIW the plot *does* change and the XHR requests to the `timeseries` API call are being made correctly, so I assume that the problem is just in the Selector rendering.

corviday:


corviday:
Arguably, it's already flagged for refactoring with #74. The repetitive code block would probably be a lot more straightforward without the whole dropdowns-passing-JSON-to-callbacks pattern.

PR #84, opened 2017-10-11
Map time select for multi-year spans, Fix #81

### LINKED ISSUE #81: Context-sensitive timestamp-selection dropdown in MapController
Currently the dropdown in MapController uses `util.timestampToTimeOfYear()` to generate the list of viewable timestamps in a file, which just displays a month, season, or "annual" based on the day of the year.

This isn't meaningful when viewing climdex data, where one file may extend over many years, leading to menus populated with dozens of items that just say "Annual". MapController needs to check whether there are timestamps from multiple years and if so, include the year in the dropdown and change the label of the dropdown to something other than "Time of Year".

corviday:


jameshiebert:
Looks pretty straightforward to me. Good work!

Issue #88, opened 2017-10-17
Help users find data that exists in the ensemble


corviday:
The `model`, `scenario`, and `variable` dropdown menus at the top of climate explorer let the user select between _all_ models, _all_ scenarios, and _all_ variables present in at least one file in the ensemble. If the user selects a combination that is not in the database (for example, `model: GISS-E2-R scenario: historical, rcp45 variable: altcddETCCDI` is not currently present in the `all_files` ensemble), climate explorer displays an error message (`No data matching selected parameters available`).

With the current ensembles, I've been seeing that message a lot. Anecdotally, it seems like over half the possible model / scenario / variable combinations a user could request are not available, at least in the `all_files` ensemble.

It might be nice if the UI indicated somehow what data exists, either by populating each selector with options available based on the other selectors (we'd need a selector hierarchy), or perhaps by indicating which options in each selector produce valid data at any given time (colouring them green or something).

On the other hand, if we intend to eventually have every variable, model, and scenario combination available, this doesn't matter much.

corviday:
Fixed by https://github.com/pacificclimate/climate-explorer-frontend/pull/98

PR #89, opened 2017-10-20
Graph and Map non multi-year-mean data

corviday:
The two major changes in this PR are:

- If the user views a dataset that is not structured to support the **Annual Cycle Graph** (monthly multi-year means) and the **Long Term Average Graph** (a timeseries of means), a **Timeseries Graph** of arbitrary resolution will be generated instead

- The map now allows the user to select the timeslice displayed separately for the raster and the isoline layers, which was necessary to support mapping different timesets at different resolutions, but will hopefully have other uses, like comparing two different time slices from the same dataset or something.

There were a lot of small adjustments needed to handle various strange edge cases and interactions. I'd appreciate it if @rod-glover and/or @jameshiebert could spare a little time to poke around the test dockers and see if there's any odd behaviour or bugs I didn't catch.

`all_files` ensemble: [MoTI portal](http://docker-dev01.pcic.uvic.ca:30210/moti), [climo portal](http://docker-dev01.pcic.uvic.ca:30210/climo), [comparison portal](http://docker-dev01.pcic.uvic.ca:30210/compare)

`all_CLIMDEX_files` ensemble: [MoTI portal](http://docker-dev01.pcic.uvic.ca:30211/moti), [climo portal](http://docker-dev01.pcic.uvic.ca:30211/climo), [comparison portal](http://docker-dev01.pcic.uvic.ca:30211/compare)

`all_downscale_files` ensemble: [MoTI portal](http://docker-dev01.pcic.uvic.ca:30212/moti), [climo portal](http://docker-dev01.pcic.uvic.ca:30212/climo), [comparison portal](http://docker-dev01.pcic.uvic.ca:30212/compare)

`all_GCM_files` ensemble: currently contains no files.


jameshiebert:
I think that the side-by-side layout of the compare page's map settings looks very good and works well. I do really like the "match time selection across variables" UI element (good idea!), but the implementation of it seems to be flaky.

E.g.:
* On load: the isolines Time of Year selector is unintuitively disabled
* Select "August" on the right and click the "match" button replaces both with annual (and enables both left and right sides)
* Select "Spring" on the left and the map will display "Spring" vs. "Summer", until you click the match button after which is maps Spring vs. Spring.
* Select "August" on the left and click match and both selectors go back to "Spring"

In short, it seems that there are a lot of state transitions here that are either unaccounted for or not properly handled.

The "Dataset Selector" on the "Annual Cycle" graph seems to be broken. I.e. it doesn't react to any changes in selection.

Selecting datasets "HadGEM2-CC" and "historical,rcp85", everything grinds to a halt. The map refuses to load and the Timeseries plot errors out.

That's all I have found so far, but I'll keep digging....

corviday:
Thanks, that's great feedback!
1. **map time matching**
It sounds like the time stamp matching is actually working as intended, but what's intended is apparently horribly unintuitive and confusing. :) The timestamp matching button is an on/off toggle, not a "copy times right now" action. When it's on, which it is by default, the isoline time selector is disabled and both map layers use the time displayed in the raster layer's selector.

What does this need? Would making it an action, not a setting (maybe two arrows, one to copy from raster to isoline, one vice versa) make more sense? Would a better tooltip explanation of the button help, like "turn on time matching" or something?

2. Annual Cycle Dataset selector
I'll look into that, that's definitely a bug.

3. HadGem2-CC historical,rcp85 
This model / scenario / variable combination doesn't exist. Isn't in the database. Is there a better way to convey this that the "**No data matching selected parameters available**" message on the plot? Would loading the background tiles of the map even if there's no data to show be an improvement?

Long term, it's a seperate issue, #88 , but it's especially bad in the all_data ensemble because each scenario exists as both "historical,rcpXX" and "historical, rcpXX", doubling the amount of non-existant data sets one could conceivably select.

jameshiebert:
re: map time matching, that makes a lot more sense given your explanation. I was definitely interpreting the button as a "copy" action rather than a toggle setting. A tooltip or some cue to the user would probably help. That said, the layers seem to update on change with the toggle is set to "off", but I haven't been able to trigger a layer update when the toggle is set to "on". The console shows "ReferenceError: dataset is not define", so there's some problem there that's not being handled at the moment.

corviday:
Some netCDF files use the same variable_id, but give it different descriptions. Here's an example:

```JSON
"rx1dayETCCDI_mon_BCCAQ_MRI-CGCM3_historical-rcp26_r1i1p1_19500116-21001216": {
    "ensemble_member": "r1i1p1",
    "timescale": "monthly",
    "start_date": "1950-01-16T00:00:00Z",
    "institution": "MRI",
    "experiment": "historical, rcp26",
    "end_date": "2100-12-16T00:00:00Z",
    "variables": {
      "rx1dayETCCDI": "Monthly Maximum 1-day Precipitation"
    },
    "model_name": null,
    "multi_year_mean": false,
    "model_id": "MRI-CGCM3"
  },
"rx1dayETCCDI_yr_BCCAQ_CCSM4_historical-rcp85_r2i1p1_19500702-21000702": {
    "ensemble_member": "r2i1p1",
    "timescale": "yearly",
    "start_date": "1950-07-02T00:00:00Z",
    "institution": "NCAR",
    "experiment": "historical, rcp85",
    "end_date": "2100-07-02T00:00:00Z",
    "variables": {
      "rx1dayETCCDI": "Annual Maximum 1-day Precipitation"
    },
    "model_name": null,
    "multi_year_mean": false,
    "model_id": "CCSM4"
  },
```

The first file uses `rx1dayETTCCDI` as a monthly variable, the second as an annual variable. They have the same variable_id, but different descriptors. This is the cause of the variable selection menu weirdness @jameshiebert noticed this morning; climate explorer currently assumes that a given `variable_id` always means the same thing.

Short term, I'll add a workaround, but I think the long term fix should probably be to ask the scientists to use separate `variable_ids` for differently scaled indices.

corviday:
I've added a temporary function that produces a single hard-coded description for the variables with multiple descriptions in different files (rx1dayETCCDI, txxETCCDI, tnxETCCDI, rx5dayETCCDI, txnETCCDI, and tnnETCCDI). This avoids the issue with trying to add the same variable to the variable dropdown menu multiple times and ending up with non-matching variable names and descriptions and undefined variables.

It doesn't allow the user to indicate _which_ rx1dayETCCDI (for example) they're interested in seeing. In practice, this means that timestamps from _both_ yearly indices and monthly indices will be available in the map, but the timeseries graph will show one or the other at random (whichever is returned first by the multimeta query).

Long term, consistent variable metadata is the preferred solution, but this workaround keeps things from going too off the rails right now.

jameshiebert:
So, I don't think that I'd agree with the assessment that there should be different variable ids for different timescales. The problem isn't that the metadata isn't consistent or complete and descriptive enough; it's that we don't have enough control on the page to disambiguate one dataset from another. I.e., we have multiple attributes in the metadata (timescale, multi_year_mean, ...) for which there is no UI control to be able to select.

I propose that we add a timescale state variable that *for now* is just hard-coded to `yearly|seasonal`, that serves as a filter on the datasets. If at some point, we are able to support monthly data in a performant manner, than we can add a control somewhere that allows the user to select different timescales.

Does this sound like a reasonable way forward?

corviday:
I could be wrong, but my understanding (from talking to @rglover) is that the majority of climdex files already treat monthly and yearly indices as separate variables; there are just a small number of files that don't and mess up the UI with multiple descriptions of their variable ids. (About sixty files across five variable ids, I think, though I don't have access to the list on Confluence right now). I believe his long term plan is to make the outliers match the majority.

rod-glover:
That is correct. There is currently a set of 60 or 80 CLIMDEX files that use the wrong variable name for an annual index. The indices in question are defined as monthly in the standard, but locally scientists also computed annual versions of these indices but did not choose a new name for them. This confused the datasets. It one of the very first things on my list today to attend to.

PR #98, opened 2017-11-22
Disable dropdown parameters that do not describe any dataset

corviday:
This change treats the parameter selectors (Model, Scenario, Year) as a left to right hierarchy, and disables any unavailable options based on selections higher in the hierarchy. The user can freely select any model. Any variables not present in that model will be disabled in the `Variable Selection` dropdown. Any emissions scenario unavailable for both the selected variable and model will be disabled in the `Emissions Scenario` dropdown.

It's still possible to specify nonexistent data parameters. For example if the user is looking at a variable and then switches models to one that doesn't have that variable, they'd get the usual error message (`No data matching selected parameters available`). But the visual cues should make it a lot easier to find data.

The selectors are a little less pretty now.

### DIFF
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -34,17 +34,24 @@ var App = React.createClass({
   },
 
   render: function () {
+    //hierarchical selection: model, then variable, then experiment
+    var modOptions = this.getMetadataItems('model_id');
+    var varOptions = this.markDisabledMetadataItems(this.getVariableIdNameArray(),
+        this.getFilteredMetadataItems('variable_id', {model_id: this.state.model_id}));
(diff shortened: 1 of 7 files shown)


corviday:
[Docker for this branch](http://docker-dev01.pcic.uvic.ca:30030/climo)

If you switch models to HadGEM2-CC, you should see that most of the variables are disabled. If you pick a variable that isn't disabled, you should see that rcp85 is the only available emissions scenario.

jameshiebert:
The functionality seems to work great!

jameshiebert:
Hey @corviday can you fix the conflicts in this PR and verify that everything still works OK?

jameshiebert:
This works great. Great work!

Issue #93, opened 2017-11-11
Selected area outline on map disappears if invalid data parameters selected


corviday:
If a user selects a set of data parameters (model, scenario, variable) which are not in the database, the map disappears, to keep users from looking at the map and thinking it shows their (invalid) selection.

If the user later selects a valid set of data parameters, the map reappears, but without the polygon showing the selected area, even though the selected area is still set in state. This gives the user no indication they're still working with a geographic selection, but they are.

Issue #91, opened 2017-11-02
Map sometimes has opacity set to 100% on overlay layers


corviday:
This bug seems to happen on initial load; changing datasets or timestamps fixes it. Need to look at calls to ncWMS and see what opacity value is being passed. If the overlay layers are opaque, context isn't visible to the user.

corviday:
Okay, it looks like when you initialize a layer and pass in a set of `params`, any` params` the layer itself understands are used to initialize the layer. It stores any parameter it doesn't understand and passes them to ncWMS on the assumption that they are WMS parameters.

After the layer is already initialized, it assumes all `params` it receives are for WMS, which makes some sense, though it does result in strange behaviours around parameter names used by both ncWMS and Leaflet.

Issue #80, opened 2017-10-03
Opening map settings menu sometimes undoes autoscaling


corviday:
**Steps to reproduce:**

- Select a temperature-based variable like tasmin or tasmax
- Using map settings menu, switch from viewing "Annual" temperatures to viewing June temperatures. The map will be mostly blank, as most June temperatures are outside the scale established by annual temperatures
- Close map settings menu and hit autoscale button. Entire map will be drawn as expected
- Open map settings menu again. It will return to the mostly blank pre-autoscale map

**Expected behaviour**
- Autoscaled map should remain autoscaled when a menu is opened

corviday:
Probable fix for the bug in question: make sure selecting a new date/time in MapController marks the previous minmax stored in state dirty.

Probably also needs to be done to prevent more subtle bugs: have autoscaling do a callback to update stored minmax as well.

corviday:
This issue exists in `master` as well as `feature/64`, and my previous guess as to its cause is incorrect.

PR #100, opened 2017-11-28
Map usability fixes

corviday:
- Fixes a bug that caused the map to refresh (losing any user changes like autoscaling) every time the map settings menu was opened ( #80 )

- Fixes a bug that caused inconsistent transparency handling of map layers, resulting in the raster layer sometimes rendered opaquely, so that a user couldn't see the context underneath. ( #91 )

- If the map is initialized while the "area" parameter was already set, the area wasn't being drawn or indicated in any way, which has now been corrected. ( #93 )


jameshiebert:
All looks good to me. Nice work!

PR #101, opened 2017-11-30
Add application-wide routing to multiple ensembles

jameshiebert:
This PR takes a look at the application routing and makes an attempt to bring together the disparate arrangement of application types and different data sets. It enables an option to use a path component to configure the ensemble name, and passes that parameter down through the application components.

[Demo docker container](http://docker-dev01.pcic.uvic.ca:30066/)

Advantages:

1. We not longer have to start _n_ different application containers to show as many different data sets/ensembles.
1. Changing between datasets and application layouts does not require a page reload. React takes care of this by only re-rendering the necessary components. So a metadata fetch is only required if the ensemble changes (not if the layout changes).

Disadvantages:

1. ensemble names are hard-coded into the navigation under this regime, which makes it *harder* (though not impossible) to dynamically configure an application with a new ensemble. (Details: one can still configure the root using the CE_BACKEND_URL environment variable, *or* navigate to /climo|compare/<ensemble_name> and get a page that shows a specific ensemble).

Current Issues:

1. Navigating from "ClimDEX comparison" to "Standard comparison"  fails to refresh the metadata. It actually errors out in DualDataController.js saying "TypeError: _underscore2.default.findWhere(...) is undefined". I have no idea why, so @corviday I'd appreciate a look (since you wrote most of that code).
1. ~~If loading the page from one of the lower path levels (e.g. /climo/all_downscale_files ) the PCIC logo fails to load. Again, I can't figure out why. If one loads the page from "/" and then navigates to the other locations, it works fine.~~
1. ~~The aethetics of the nav buttons is kind of... odd. It works OK, but seems to stand out and not fit with the rest of the page. Anyone willing to help?~~

### DIFF
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -24,12 +24,6 @@ var App = React.createClass({
    * Includes: - model_id - variable_id - experiment
    */
 
-  getDefaultProps: function () {
-    return {
-      ensemble_name: CE_ENSEMBLE_NAME
-    };
-  },
-
   mixins: [AppMixin],
 
   //This filter controls which datasets are available for viewing on this portal;
@@ -64,6 +58,7 @@ var App = React.createClass({
           </Col>
           <Col lg={6}>
             <DataController
+              ensemble_name={this.state.ensemble_name}      
               model_id={this.state.model_id}
               variable_id={this.state.variable_id}
               experiment={this.state.experiment}
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -24,6 +24,12 @@ var App = React.createClass({
    * Includes: - model_id - variable_id - experiment
    */
 
+  getDefaultProps: function () {
+    return {
+      ensemble_name: CE_ENSEMBLE_NAME
+    };
+  },
+
   mixins: [AppMixin],
 
   //This filter controls which datasets are available for viewing on this portal;
--- a/src/components/AppController/AppController.js
+++ b/src/components/AppController/AppController.js
@@ -58,11 +58,8 @@ var App = React.createClass({
           </Col>
           <Col lg={6}>
             <DataController
-              ensemble_name={this.state.ensemble_name}      
-              model_id={this.state.model_id}
-              variable_id={this.state.variable_id}
-              experiment={this.state.experiment}
-              area={this.state.area}
+              // ensemble_name, model_id, variable_id, experiment, area
+              {...this.state}
               meta = {this.getfilteredMeta()}
             />
           </Col>
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -18,15 +18,21 @@ import {timestampToYear} from '../../core/util';
 
 var AppMixin = {
   getInitialState: function () {
+    var ensemble_name = (typeof this.props.params !== 'undefined') ?
+        this.props.params.ensemble_name : ((typeof this.props.ensemble_name !== 'undefined') ?
+                                          this.props.ensemble_name : CE_ENSEMBLE_NAME);
     return {
-      ensemble_name: this.props.params.ensemble_name || this.props.ensemble_name || CE_ENSEMBLE_NAME,
+      ensemble_name: ensemble_name,
       meta: [],
     };
   },
 
   componentWillReceiveProps: function(nextProps) {
+    var ensemble_name = (typeof nextProps.params !== 'undefined') ?
+        nextProps.params.ensemble_name : ((typeof nextProps.ensemble_name !== 'undefined') ?
+                                          nextProps.ensemble_name : CE_ENSEMBLE_NAME);
     this.setState({
-      ensemble_name: nextProps.params.ensemble_name || nextProps.ensemble_name || CE_ENSEMBLE_NAME,
+      ensemble_name: ensemble_name,
     });
   },
 
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -24,7 +24,17 @@ var AppMixin = {
     };
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    this.setState({
+      ensemble_name: nextProps.params.ensemble_name || nextProps.ensemble_name || CE_ENSEMBLE_NAME,
+    });
+  },
+
   componentDidMount: function () {
+    this.updateMetadata();
+  },
+
+  updateMetadata: function () {
     var models = [];
     var vars;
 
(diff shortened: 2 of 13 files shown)


rod-glover:
I'll work on the nav button aesthetics. As well as see what I can figure out about the logo not loading.

src/components/AppController/AppController.js
```diff
@@ -58,6 +58,7 @@ var App = React.createClass({
           </Col>
           <Col lg={6}>
             <DataController
+              ensemble_name={this.state.ensemble_name}
```
rod-glover: You can simplify passing state to props using [spread attributes](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes):
```
<DataController
  {...this.state}
  meta = {this.getfilteredMeta()}
/>
```
But given that definition of `this.state` is hidden in `AppMixin`, you might decide it's clearer to be explicit about the contents of the state variable you are passing. Another reason to dislike mixins, especially as used in `AppMixin`.

src/components/AppMixin/AppMixin.js
```diff
@@ -48,17 +65,38 @@ var AppMixin = {
...
-         });
+        var newstate = _.mapObject(_.pick(this.state, 'model_id', 'variable_id', 'experiment'), (val, key) => {
+            return _.contains(_.pluck(models, key), val) ? val : models[0][key];
         });
+
+        this.setState({
+          meta: models,
+          model_id: newstate['model_id'],
+          variable_id: newstate['variable_id'],
+          experiment: newstate['experiment']
```
rod-glover: ... replace this with just 
```
this.setState({
    meta: models,
    model_id,
    variable_id,
    experiment
});
```
Ain't ES6 wonderful?

src/components/AppMixin/AppMixin.js
```diff
@@ -48,17 +65,38 @@ var AppMixin = {
...
+        // is not available from the current metadata, simply use
+        // the first available
 
-         this.setState({
-         meta: models,
-         model_id: models[0].model_id,
-         variable_id: models[0].variable_id,
-         experiment: models[0].experiment,
-         });
+        var newstate = _.mapObject(_.pick(this.state, 'model_id', 'variable_id', 'experiment'), (val, key) => {
```
rod-glover: Could use destructuring assignment here:
```
const {model_id, variable_id, experiment} = ...
```
and ...

src/components/Header/Header.css
```diff
@@ -3,3 +3,7 @@
     padding: 15px 0px;
     height: 100%;
 }
+
+.active {
+    color: green;
+}
```
rod-glover: Remove this if you adopt `Nav`, below.

src/index.js
```diff
@@ -31,9 +31,9 @@ var App = React.createClass({
 render((
   <Router history={browserHistory}>
     <Route path='/' component={App}>
-      <Route path='moti' component={MotiController} />
-      <Route path='climo' component={AppController} />
-      <Route path='compare' component={DualController} />
+      <Route path='/moti' component={MotiController} />
+      <Route path='/climo/:ensemble_name' component={AppController} />
+      <Route path='/compare/:ensemble_name' component={DualController} />
```
rod-glover: Nice

rod-glover:
Regarding item 2 (loading), when I browse directly to [climo/all_downscale_files](http://docker-dev01.pcic.uvic.ca:30066/climo/all_downscale_files), I get an entirely blank page. Nada. Zip.

src/index.js
rod-glover: But should you be [using `Router` components, not `Route`](https://stackoverflow.com/a/41555097/1858846) here?

rod-glover:
See comment regarding `Router` vs. `Route` above.

src/components/AppMixin/AppMixin.js
```diff
@@ -18,18 +18,35 @@ import {timestampToYear} from '../../core/util';
 
 var AppMixin = {
   getInitialState: function () {
+    var ensemble_name = (typeof this.props.params !== 'undefined') ?
+        this.props.params.ensemble_name : ((typeof this.props.ensemble_name !== 'undefined') ?
+                                          this.props.ensemble_name : CE_ENSEMBLE_NAME);
```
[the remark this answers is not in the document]
jameshiebert: Thanks, I knew there was a better way to do that!

src/components/AppMixin/AppMixin.js
```diff
@@ -18,18 +18,35 @@ import {timestampToYear} from '../../core/util';
...
     return {
+      ensemble_name: ensemble_name,
       meta: [],
     };
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    var ensemble_name = (typeof nextProps.params !== 'undefined') ?
+        nextProps.params.ensemble_name : ((typeof nextProps.ensemble_name !== 'undefined') ?
+                                          nextProps.ensemble_name : CE_ENSEMBLE_NAME);
```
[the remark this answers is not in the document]
jameshiebert: Hm, there *is* a logic error there, but not the one you think. The intent is for there to be three places to look (in order of preference):
1. props.params.ensemble_name (comes from the path)
1. props.params (comes from an instantiated component)
1. CE_ENSEMBLE_NAME (global config)

With the syntax changes and DRYing it up, this should become both clear and error-free :)

src/components/Header/Header.js
```diff
@@ -21,8 +22,22 @@ class Header extends Component {
                 />
               </a>
             </Col>
-            <Col lg={4} />
-            <Col lg={4} />
+            <Col lg={8}>
+              <Row>
+                <Link to="/climo/all_downscale_files" activeClassName="active">
```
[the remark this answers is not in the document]
jameshiebert: Perfect, I totally missed the Nav component. Great suggestion!

src/index.js
jameshiebert: Huh? I *am* using the Router component...

rod-glover: Sorry, I must have misread the code.

rod-glover: I will continue researching this.

jameshiebert: Yeah, if you look at the page errors, you'll see:

```
Loading failed for the <script> with source “http://localhost:8080/climo/bundle.js”.
```

So when you do an initial page load at one of the child paths, it tries to load all of the resources in the index.html (e.g. bundle.js (which is why you see nothing), all of the leaflet stuff, etc.) *from that path* and fails. But if you do the initial page load at the root, it gets all of those resources, and then never has to reload the page on navigation, because React takes care of all of the navigation through re-rendering components.

But yeah, if you have any suggestions, I'd appreciate them.

src/components/Header/Header.js
```diff
@@ -40,6 +49,6 @@ class Header extends Component {
       </div>
     );
   }
-}
+});
```
rod-glover: This is fairly canonical, but I have some comments:

1. The navigation elements can and should be separated out into a component independent of Header. Call it, say, `Navigation`, and invoke it here. It will encapsulate all the nav stuff, including the state and handlers that make the active selection behave. This is very much in the spirit of React programming. Never hesitate to create a component that can encapsulate and isolate complexity, particularly if that includes state. State should always be as far down the hierarchy as possible.
2. Make `updateNav()` and `handleSelect()` _methods of the class_ (your new `Navigation` class). It looks odd to have them inside `render()`; they look as if they are (although they are not, actually) modifying state inside `render`, which is verboten.

rod-glover: Another thought. I think this won't correct set the navigation when you drive directly to a route via URL. Is there a way you can use the `activeHref` property of `Nav` and always set it from the current path? I think that ought to be straightforward but I don't know it for sure. It sounds more robust, and it saves you having to manage the state explicitly in the `Navigation` component -- the router will do that for you.

jameshiebert: Great suggestions, though I think that I'm out of time on this issue. Going to write up an enhancement issue, point to this discussion and defer to the work to our next round of revisions.

PR #104, opened 2017-12-08
Update comparison variable when switching ensembles

corviday:
Added a check to `AppMixin.updateMetadata()` that checks when switching ensembles to see whether a comparison variable (used by the `/compare` portal) is defined and if so, whether it refers to a variable not present in the new ensemble, updating it if so.


corviday:
Closing for now, @rod-glover mentioned a better way to structure this.

PR #110, opened 2017-12-19
Fix #103, update comparand on ensemble switch

### LINKED ISSUE #103: Change comparison variable after ensemble switch
@jameshiebert[ says](https://github.com/pacificclimate/climate-explorer-frontend/pull/101):

> Navigating from "ClimDEX comparison" to "Standard comparison" fails to refresh the metadata. It actually errors out in DualDataController.js saying "TypeError: _underscore2.default.findWhere(...) is undefined".

Main variable on all portals is set by `AppMixin.componentDidMount()`. Since DualController is the only portal that has a comparison variable, the comparison variable is set separately from other portal state by `DualController.componentDidUpdate()`. That function needs to be updated to handle the case where the user changes ensembles and the comparison variable is already initialized, but to a variable not present in the new ensemble.

corviday:
Updates secondary variable in comparison portal after switching ensembles. Checks to make sure _both_ variables are valid before trying to load data from the backend.

### DIFF
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -98,7 +98,7 @@ var AppMixin = {
 
   componentDidUpdate: function(nextProps, nextState) {
     // The metadata needs to be updated if the ensemble has changed
-    if (nextState.ensemble_name !== this.state.ensmeble_name) {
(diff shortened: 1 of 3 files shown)


rod-glover:
LGTM

PR #114, opened 2018-01-29
Update internal portal routing URLs

corviday:
Fix the URLs from the portal tabs at the top of the page. Upgrading react-router to version 3 seems to have changed how they work.


corviday:
The issue addressed by this pull request has been resolved.

jameshiebert:
Thanks for catching it, though.

PR #117, opened 2018-02-06
Add multi-model context graph to main portal

corviday:
Adds a multiple-model "spaghetti plot" when viewing datasets with precalculated multi-year-means on the /climo portal. The graph - at least on my workstation - takes about ten seconds to load, due to querying the backend separately for each model. It does seem to delay generation of other graphs, but only very slightly. I'll get a docker running so we can judge that.

A version of this graph for datasets that lack precalculated multi year means is not included in this pull request; it was too slow.

### DIFF
--- a/src/components/DataController/DataController.js
+++ b/src/components/DataController/DataController.js
@@ -23,9 +23,6 @@
  * If the selected dataset is not a multi year mean:
  *  - a Time Series Datagraph showing each time point available.
  *
- *  - a Model Context datagraph similar to the Time Series Datagraph,
- *    but showing data for each model.
- *
  * A Data Table viewer component showing statistical information for each
  * climatology period or timeseries is also generated. 
  *******************************************************************/
@@ -102,7 +99,6 @@ var DataController = createReactClass({
     else {
       this.loadTimeseriesGraph(props);
       this.loadDataTable(props, {timeidx: 0, timescale: "yearly"});
-      this.loadContextGraph(props);
     }
   },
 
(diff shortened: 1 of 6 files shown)


jameshiebert:
Looks awesome! One small suggestion: I'd recommend not reloading the graph data when the model gets changed. The data is all the same (hence we shouldn't need to wait). You would just need to re-render the component with a different active model. That should be pretty fast.

jameshiebert:
PS I have a docker container up [here](http://docker-dev01.pcic.uvic.ca:30088).

corviday:
Good idea, that should be pretty easy.

jameshiebert:
Great, that works a lot faster! There's still a flicker of 2-3 re-renders as events come in *after* the emphasize current model. It's not a *huge* deal, but you may want to add a ``shouldComponentUpdate`` lifecycle function to the component.

Other than that, it looks good to me. @rod-glover do you want to review for code quality?

corviday:
Unfortunately, at present the graphs are not individual components; they're completely generated and managed by `DataController`. `DataController` has a `shouldComponentUpdate()`, but it returns true when any of the three graphs (or the stats table) finishes fetching and parsing data and makes a `setState()` call.. So all three graphs and the stats table are rerendered any time one of them updates, which is what is causing the extra flickers in this case.

Long term, the right way to go is presumably to encapsulate each graph as a separate component with its own `shouldComponentUpdate()`, which would have benefits like fewer flickers and reduced complexity and repetitiveness of the data controllers.

jameshiebert:
Uggh, I hadn't realized that they are all glommed together in one component. Yes, we absolutely need to separate them, especially since there are three new plots on our TODO list.

But that's for another day, not this PR. I'll file an issue on that.

src/components/DataController/DataController.js
```diff
@@ -87,6 +95,7 @@ var DataController = createReactClass({
       this.loadAnnualCycleGraph(props);
       this.loadLongTermAverageGraph(props);
       this.loadDataTable(props);
+      this.loadContextGraph(props);
```
rod-glover: These method names are confusing in several ways:

- they don't load a graph, they load the data for a graph
- they set state with an object called "graph", but the state property name is "xxxData"
- "graph" here actually means an object containing data and C3 options that is used (via DataGraph to render a C3 graph)

Strongly suggest that identifiers are precise about what they do or contain, and do not use confusing abridgements of meaning as here. This occurs in many places in this code. Kindly rename these so that their names reflect their use and meaning.

src/components/DataController/DataController.js
```diff
@@ -17,6 +17,9 @@
  *  - a Long Term Average Datagraph showing the mean of each climatology
```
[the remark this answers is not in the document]
rod-glover: Also, when you adopt React Bootstrap Tabs, you will not have to separate the creation of tab labels and tab panels. They are unified, and that will make writing the conditional code considerably simpler.

src/components/DataController/DataController.js
```diff
@@ -296,14 +328,90 @@ var DataController = createReactClass({
...
+      }).catch(error => {
+        this.displayError(error, this.setContextGraphNoDataMessage);
+      });
+    }
   },
 
   render: function () {
-    var longTermAverageData = this.state.longTermAverageData ? this.state.longTermAverageData : this.blankGraph;
-    var annualCycleData = this.state.annualCycleData ? this.state.annualCycleData : this.blankGraph;
     var statsData = this.state.statsData ? this.state.statsData : [];
```
rod-glover: In the case that there is no `statsData`, this will return a new empty array each time, which will likely force unnecessary re-renders in whatever components are passed this state. Suggest you substitute a fixed empty array defined on `this` or even on the class (a static variable), as is done for the other data items.

src/core/chart.js
```diff
@@ -790,8 +798,56 @@ var fadeSeriesByRank = function (graph, ranker) {
   return graph;
 };
 
+/*
+ * Post-processing graph function that removes data series from the legend.
```
rod-glover: Nice.

corviday:
Most of the requested changes apply to code that predates this PR and affect all the Data Manager components. Would we like to make these changes to all three data managers as part of this PR?

jameshiebert:
>  have decided that refactoring DataControllers belongs in a separate PR.

Agreed. @rod-glover feel free to merge this whenever you're satisfied with that your comments on the *changes* in this PR have been addressed (or delegated to future work).

corviday:
I think I got all the issues that were confined to this controller:
- [ ] add `PropType` for `contextMeta`
- [ ] static array for empty Stats Table to prevent gratuitous rerendering
- [ ] add a `value` prop to the time selector on the Long Term Average Graph so it displays current time

Issues that affect all three DataControllers and will be handled seperately:
- [ ] Unclear function and variable names (#130)
- [ ] Encapsulating (and reusing across controllers!) individual graphs (#119)

jameshiebert:
Awesome, thanks guys! I'll update the beta.

PR #120, opened 2018-02-06
Fix DualDataController query bugs

corviday:
Fix two minor DualDataController bugs discovered while testing the react refactor PR.

1. DualDataController was sometimes sending off queries to the `data` API without an ensemble parameter. Added that.

2. [Recently switched](https://github.com/pacificclimate/climate-explorer-frontend/pull/116) to using the props passed to the query instead of props stored on the controller for determining whether the data being queried was a multi-year-mean, which led to an issue in `DualDataController`. `DualDataController` "fakes" props for queries on the secondary variable; it parsimoniously mocks up props containing only attributes needed to run the query. The fake props didn't contain enough information to determine whether the comparand dataset was a multi-year mean, which was preventing the secondary variable (`comparand`) from loading on some graphs. Added the missing attributes to the mockup props.

### DIFF
--- a/src/components/DualDataController/DualDataController.js
+++ b/src/components/DualDataController/DualDataController.js
@@ -247,7 +247,7 @@ var DualDataController = createReactClass({
 
     var timeseriesPromises = [];
     
-    var variableTimeseriesParams = {variable_id: props.variable_id, area: props.area};
+    var variableTimeseriesParams = _.pick(props, "variable_id", "area", "meta");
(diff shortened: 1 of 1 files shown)


corviday:
A future change, maybe: write a general "create fake props for comparand variable queries" function? This is the second bug I've introduced relating to leaving something essential out of the mockup props; it would probably help if there was one piece of code that did that, not four.

corviday:
After discussion with Rod, I have decided to make the change referenced in the previous comment part of this PR.

PR #132, opened 2018-02-10
Drop "modtime" from file metadata used inside Climate Explorer

corviday:
The climate explorer frontend doesn't need to know about modtime of queries run to the backend.

Functions that look for matching datasets based on metadata attributes or write human-friendly descriptions of data based on metadata attributes were taking modtime into account, with unhelpful results.


jameshiebert:
Perfect. Thanks for the fast diagnosis and fix.

PR #136, opened 2018-02-16
Add currently running version number and commitish to footer

corviday:
screenshot from 2018-02-16 12-19-03
This is a little awkward. It uses a shell script run at startup to store a current version string in the environment variable `CE_CURRENT_VERSION`, which is then imported along with the other environment variables like `CE_BACKEND_URL`. 

I didn't find a good path forward for a pure javascript solution. simple-git wasn't playing well with our version of Node, and nodegit is a very powerful set of bindings to libgit, but people have reported having a hard time building it on Travis, which gave me pause. 

I'm open to suggestions for alternate approaches.

### DIFF
--- a/generate-commitish.sh
+++ b/generate-commitish.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+#Generates a commitish string for climate explorer and stores it in 
+# the CE_CURRENT_VERSION environment variable.
+
+VERSIONTAG="$(git describe --tags --abbrev=0)"
+BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+COMMITSHA="$(git log -1 --format=%h)"
+
+export CE_CURRENT_VERSION="Climate Explorer $VERSIONTAG ($BRANCH: $COMMITSHA)"
+echo "generate-commitish.sh: Current version is ${CE_CURRENT_VERSION}"
+
--- a/generate-commitish.sh
+++ b/generate-commitish.sh
@@ -6,6 +6,4 @@ VERSIONTAG="$(git describe --tags --abbrev=0)"
 BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 COMMITSHA="$(git log -1 --format=%h)"
 
-export CE_CURRENT_VERSION="Climate Explorer $VERSIONTAG ($BRANCH: $COMMITSHA)"
-echo "generate-commitish.sh: Current version is ${CE_CURRENT_VERSION}"
-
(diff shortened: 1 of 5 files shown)


generate-commitish.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#Generates a commitish string for climate explorer and stores it in 
+# the CE_CURRENT_VERSION environment variable.
+
+VERSIONTAG="$(git describe --tags --abbrev=0)"
+BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+COMMITSHA="$(git log -1 --format=%h)"
+
+export CE_CURRENT_VERSION="Climate Explorer $VERSIONTAG ($BRANCH: $COMMITSHA)"
```
jameshiebert: Drop the export here statement here, and drop the "Climate Explorer" prefix from the version information.
Just output the info we need straight-up:

```bash
echo  $VERSIONTAG ($BRANCH: $COMMITSHA)
```

generate-commitish.sh
```diff
@@ -0,0 +1,11 @@
...
+echo "generate-commitish.sh: Current version is ${CE_CURRENT_VERSION}"
```
jameshiebert: Drop this altogether.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "prestart": "if [ ! -f ./variable-options.yaml ]; then touch ./variable-options.yaml; fi",
-    "start": "webpack-dev-server --host 0.0.0.0 --history-api-fallback",
+    "start": ". ./generate-commitish.sh; webpack-dev-server --host 0.0.0.0 --history-api-fallback",
```
jameshiebert: Rather than having the script set an environment variable (side effects == bad), use the output of the script to set the variable on the run line. E.g.:

```
CE_CURRENT_VERSION=$(./generate-commitish.sh) webpack-dev-server --host 0.0.0.0 --history-api-fallback
```

generate-commitish.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#Generates a commitish string for climate explorer and stores it in 
+# the CE_CURRENT_VERSION environment variable.
+
+VERSIONTAG="$(git describe --tags --abbrev=0)"
```
jameshiebert: If we intend to use git, we'll have to ensure that git is included in whatever Docker base we have.

Turns out that it is (but that actually surprises me):

```
hiebert@docker-dev01:~$ docker run -it node:latest git --version
git version 2.1.4
```

PR #137, opened 2018-02-19
Refactor graph code into components

### LINKED ISSUE #119: Split each plot into its own component
From the conversation in PR #117:

> Unfortunately, at present the graphs are not individual components; they're completely generated and managed by ``DataController``. ``DataController`` has a ``shouldComponentUpdate()``, but it returns true when any of the three graphs (or the stats table) finishes fetching and parsing data and makes a setState() call.. So all three graphs and the stats table are rerendered any time one of them updates, which is what is causing the extra flickers in this case.

Agreed. We need to split the plots into each being their own component, accepting only what data they need and rendering as required.

### LINKED ISSUE #130: Confusing names for methods and variables related to graphs
In DataController (and its cousins), the naming of methods and variables related to graphs is confusing. 

Specifically, there is a set of methods called `load*Graph` with associated internal variables and external state they update with confusing naming:

- they don't load a graph, they load the data for a graph
- they set state with an object called "graph", but the state item is `state.*Data`
- "graph" here actually means an object containing data and C3 options that is used (via `DataGraph`) to render a C3 graph - suggest referring to these objects as `*GraphSpec`.

Strongly suggest that identifiers are precise about what they do or contain, and do not use confusing abridgements of meaning as here. This occurs in many places in this code. Kindly rename these so that their names reflect their use and meaning.

rod-glover:
Resolves #119 
Resolves #130 

Apologies for another large refactoring. Fortunately, this one is simpler in structure.

This refactoring creates components for each type of graph presented in the UI. See [graphs/README.md](https://github.com/pacificclimate/climate-explorer-frontend/blob/issue119-graph-components/src/components/graphs/README.md) for a bit more detail on this.

The original code resided in `DataController` and `DualDataController`, and for the most part constructed the graphs procedurally. That code has now been simplified, refactored and redistributed to reusable components. A few remarks on the new code structure:

- Data controllers are now much simpler and use entirely declarative code for graphs.
- Graph components group all code for a single graph type, and only that code, into one place.
- Lifecycle event handling is simpler, more localized, and [according to canon](https://pcic.uvic.ca/confluence/display/CSG/React+Lifecycle#ReactLifecycle-FetchingDataandUpdatingStateintheLifecycle).

The alert reviewer will note the great similarity between the 4 components `AnnualCycleGraph`, `ContextGraph`, `LongTermAveragesGraph`, and `TimeSeriesGraph`. These similarities could be wrapped up in a HOC, which would reduce redundancy but take more time. I have not done that (yet). I can either do that or add an issue.


corviday:
I vote for reviewing and merging this, and making HOC-ification a separate PR, so that I can use the good work you've done here in the extreme precipitation portal I'm working on now.

src/components/graphs/LongTermAveragesGraph/LongTermAveragesGraph.js
```diff
@@ -0,0 +1,160 @@
...
+import {
+  timeKeyToResolutionIndex,
+  validateLongTermAverageData,
+} from '../../../core/util';
+import { getData } from '../../../data-services/ce-backend';
+import { exportDataToWorksheet } from '../../../core/export';
+
+// This component renders a graph, over actual time points (as opposed to the
+// "average year" time points of an annual cycle graph), of spatially averaged
+// values of long-term average data for the selected model, variable, and
```
corviday: This description is a little confusing to me. The Long Term Average graph shows a series of averages for subsequent (and sometimes overlapping) climatological periods. I wouldn't say it represents "actual time points" any more than the annual cycle graph, which also shows average values from climatological periods.

In other words, if your climatological mean data looks like this:
```
          |2010 - 2040 | 2040 - 2070|
------------------------------------
January   | 1          | 3          |
February  | 2          | 4          |
```

The Annual Cycle Graph graphs [1, 2] or [3, 4] depending on which climatology is set on its dropdown, but the Long Term Average Graph graphs [1, 3] or [2, 4] depending on which month is set on its dropdown.

The only "actual time" graph, which is used only for datasets without climatological means, is the TimeSeriesGraph.

corviday:
Testing seems great! It behaves identically to the pre-refactor version, as far as I can tell.

src/components/graphs/ContextGraph/ContextGraph.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      // Note omission of model_id: This graph doesn't need to reload if model
+      // doesn't change. See `componentWillReceiveProps`.
+      prevProps.variable_id !== this.props.variable_id ||
+      prevProps.experiment !== this.props.experiment ||
+      prevProps.contextMeta !== this.props.contextMeta ||  // TODO: Necessary?
```
corviday: I think currently the only circumstance in which `contextMeta` changes would be when the ensemble does, since `contextMeta` is currently just "all available dataset metadata" - the full results of the `multimeta` query.

... which, now that I think about it, doesn't seem ideal. No reason to pass` DataController` a bunch of file metadata it isn't going to use. contextMeta should probably be filtered to all datasets with the selected` variable_id` and `experiment` by `AppController`, and change when the user picks different values for `variable_id` or `experiment`. In that case, it would change when `ensemble`, `variable_id`, or `experiment` did, but not otherwise. That change is probably outside the scope of this PR, though.

rod-glover: Added a new issue for this.

Issue #126, opened 2018-02-07
Replace MapController with separate controllers specialized for each app


rod-glover:
Currently `MapController` serves all apps. This is already unwieldy, and will be come untenable when other apps (e.g., streamflow) are added. Instead, it would be better to build specialized controllers for each type of app from the components that `MapController` has now been broken up into.

DO NOT do this until https://github.com/pacificclimate/climate-explorer-frontend/issues/118 has been completed.

Consider whether this is better done before or after https://github.com/pacificclimate/climate-explorer-frontend/issues/125. After, probably.

Issue #118, opened 2018-02-06
Refactor "datasets"


rod-glover:

corviday:
For what it's worth, when we [ran into this issue in DataController](https://github.com/pacificclimate/climate-explorer-frontend/pull/73), we ended up naming the "**model+emissions+variable+start+end+run combination, which describes a collection of data that may or may not be spread across multiple files**" thing an "**instance**." I'd intended to rename MapController's **dataset** variables to **instance** variables someday but we all know what intending to do a thing someday is worth. :)

Like MapController, DataControllers normally just represents an "instance" as just `period` and `run` in state, since it gets model, emissions, and variable (explicitly, unlike MapController) from its parent as props. 

I'm not in any way advocating in favour of the name "instance," which is pretty generic, just pointing out possibly useful parallels.

corviday:
"Datasets" are now named "DataSpecs". But they still need refactoring, alas.

PR #141, opened 2018-03-03
Extreme Precipitation Portal

corviday:
Adds a new portal for understanding precipitation extremes, `PrecipitationController`. This portal uses the existing `DualDataController`, but one of the two variables is always precipitation. Adds a variable response graph to `DualDataController`. Adds a new precipitation-centric map controller, `PrecipMapController`, which always renders logarithmically scaled color mappings, and shows precipitation as annotated isolines.

Changes made in support of the new functionality that affect the whole codebase:
1. Annual Cycle graphs will now display any combination of resolutions; previously they explicitly required at least one monthly dataset.
2. Emissions scenarios are converted to a standard format when processing the multimeta query, and un-standardized when passed to the backend via queries.
3. ` MapController` was split into three components: `SingleMapController`,` DualMapController`, and `PrecipMapController`. Common data-handling functionality was encapsulated in map-helpers.js, but there's still a lot of repeated code for wiring things up. More work will need to be done to really get the `MapControllers` into as nice a shape as the graphs are now. See #126 
4. The Variable Response Graph is available to all comparison portals; it's part of `DualDataController`.
-----------------------------------------------
I finally completed the `dataset `/ `instance` split. Once upon a time, the parameters input by the user (variable, experiment, emissions scenario, start date, end date, and run) always uniquely described a single file containing all the data they were interested in, which we called a dataset. 

When we began separating different resolutions in different netCDF files (monthly, seasonal, annual), in accordance with the CF Standard, the two concepts "the set of parameters a user has selected" and "a file that contains data the user wants" no longer had a 1 to 1 relationship. This led to lots of variables and functions named `datasetWhatever` that might, confusingly, refer to either either "a specific file with data in it" or "a set of parameters uniquely describing the data a user would like to see, but the data may be spread across up to three files."

We've been slowing converting every time the word "dataset" means "a set of parameters describing the data a user wants" to the word "instance" instead, as we work on each file affected. The only place that confusion was left was in MapController and its children, and I think I got it all cleaned up while splitting MapController into three components.

**Instance** = the set of parameters describing the data a user is exploring (variable, period, run, model, and emissions scenario)
**Dataset** = a specific file with data in it

We still have some really kludgy handling around the way instances work - they're frequently serialized and passed between components as strings the way filename strings used to be - but that's for another PR. See https://github.com/pacificclimate/climate-explorer-frontend/issues/118 or https://github.com/pacificclimate/climate-explorer-frontend/issues/74 , but the variable names should be less confusing.

-----------------------------------------------
There is a test docker [here](http://docker2.pcic.uvic.ca:30302/#/precipitation/test_precipitation), but I hope to have a better ensemble with more variables for it soon. It is currently running against an ncWMS instance that only serves precipitation data, so if you look at non-precipitation portals, you won't see any maps. Should still be possible to evaluate the new Variable Response Graph, though.

The test ensemble also does not contain any annual climdex data for the rcp85 scenario. This is not a frontend bug.


corviday:
I did something "creative" with a git rebase, apparently; the first two commits on the PR are actually already part of master.

jameshiebert:
I'd appreciate it if you can do another proper rebase to get rid of those first two commits. You should be able to do an interactive rebase (``git rebase -i``) and select the "drop" option for the first two commits.

src/components/MapController/DualMapController/DualMapController.js
```diff
@@ -123,123 +131,77 @@ export default class MapController extends React.Component {
...
-          }
-        }
+    // and timestamp to the viewer component DataMap.
+    // The variable and the comparand may have different available timestamps, but will
+    // (aside from variable_id) display the same instance.
+
+    const { start_date, end_date, ensemble_member } = instance;
+    
+    let rasterParamsPromise = getRasterParamsPromise(instance, props.meta);
+    let isolineParamsPromise = getIsolineParamsPromise(instance, props.comparandMeta);
```
rod-glover: The following little essay has impact on all the map controllers. I'm putting it here in one place so that I don't distribute and/or repeat the explanations.

Also, I'm not stuck on this being the absolute right way to go, especially in view of the likely need to bind in additional data to the downstream processing functions. But it is worth knowing and considering. I invite your thoughts.

### Principle

It's a good idea to separate things that are inherently promises from those that only coincidentally are. In this case,  `rasterParams` (and `isolinesParams` and `annotatedParams`) are only coincidentally promises, since they are computed from something originally fetched asynchronously. The original fetch is inherently a promise, but the computations downstream of that are only coincidentally promises via the `.then()` chaining. 

A common and helpful practice is to formulate such promise chains as follows:

```
const resultPromise = fetchThing()  // fetchThing returns an inherent promise
  .then(doSomething)
  .then(doSomethingElse)
  .then(doYetAnotherThing);
```

The processing functions themselves do not return promises, they just process values. This separates inherent promises (the result of `fetchThing` and the `then`s) from the downstream processing. 

### Practice

The effect and the point of all this is probably best illustrated with a real example from this code: `getRasterParamsPromise` would be reformulated to

```
rasterParamsPromise = fetchTimeMetadata(datafiles)
   .then(extractTimes)
   .then(makeScalarParams)
   .then(selectRasterPalette)
```

where 
* `fetchTimeMetadata` returns an inherent promise, the `Promise.all()` for the set of metadata fetches in `getTimesPromise`
* `extractTimes` is the anonymous arrow function in `getTimesPromise`
* `makeScalarParams` is the body of `getScalarParamsPromise`
* `selectRasterPalette` is the body of `getRasterParamsPromise`

### Benefits

1. The chaining of `then`s, combined with descriptive function names, makes the processing close to self-documenting, with all the high-level info available in one place at a single glance.
2. The chained functions can be reused in other similar chains. E.g., for the equivalent chains for `getIsolineParamsPromise` and `getAnnotatedParamsPromise`. You've done a nice job factoring these already, but it can get even clearer if you separate promise from non-promise. The code gets slightly more vebose and repetitious (the promise chains are very similar) but it increases clarity.
3. Promises are returned only by functions that inherently must: the `fetch` and the `then`s. The rest are promise-agnostic. This matters because you can create unnecessary tangles (and unrefactorable code) when things that don't _have to be_ concerned with promises _are_. This is an aspect separation of concerns in the context of promises.

### Downsides

1. Verbosity. (Compensated by increased clarity.)
2. The downstream chained functions sometimes need information (values) that they don't get from their chaining inputs. This can be accomplished by partially applying (binding) functions taking the full set of params, yielding a function that takes just the chaining inputs. Not fun, but not really hard either. (You can also pass these extras along the `then` chain, but that usually turns out to be a very bad idea. It would be in this case.)

src/components/MapController/map-helpers.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+// predicate that detects whether a timestamp index is a 0th index:
+// either January, winter, or the first year in a file. 
+const is0thIndex = timestamp => (JSON.parse(timestamp).timeidx == 0);
+
+/********************************************************************
+ * 1. WMS parameter generating functions
+ ********************************************************************/
+
```
rod-glover: See comment about promise chaining. It will change what's here if we decide to do it.

src/components/MapController/DualMapController/DualMapController.js
```diff
@@ -123,123 +131,77 @@ export default class MapController extends React.Component {
...
+      if(!newVariable) {
+        rasterParams.palette = this.state.raster.palette;
+        rasterParams.logscale = this.state.raster.logscale;
+      }
+      if(!newComparand) {
+        isolineParams.palette = this.state.isoline.palette;
+        isolineParams.logscale = this.state.isoline.logscale;
+      }
+      
+      if(isolineParams.palette == 'x-Occam' &&
```
rod-glover: Use `===`; avoid `==` ([it's type-unsafe](https://www.codementor.io/thinker3197/avoid-common-pitfalls-in-javascript-u0hzz2nfw)).

If you make one change to your JS habits, it should be this. `==` is Satan's tool.

src/components/AppController/AppController.js
```diff
@@ -13,7 +13,7 @@ import React from 'react';
 import createReactClass from 'create-react-class';
 import { Grid, Row, Col } from 'react-bootstrap';
 
-import MapController from '../MapController';
+import SingleMapController from '../MapController/SingleMapController';
```
rod-glover: Because variants of different controller components are proliferating, and somewhat confusingly named, I'd like to introduce some grouping and naming conventions. Instead of putting specific suggestions all over this PR, I'll say them just once here, with application in many places.:

1. Put all app controllers in a subdirectory `./app-controllers`.
2. Put all data controllers in a subdirectory `./data-controllers`
3. Rename the subdirectory `./MapController` to `./map-controllers`.
2. Use consistent naming for data controller and map controller components. Name pattern is `<Portal><Thing>Controller`, where `<Portal> = Single | Dual | Precip` and `<Thing> = App | Map | Data`.
5. Where applicable (in most if not all of the app and data controllers), do the following:
   1. Just do `export default createReactClass({ ...`. No need to create an intervening variable (often misnamed, e.g.,  `App`) that is immediately exported.
   2. Set `displayName` to the component's name (according to convention above). There are several places where `displayName` is wrong and will show the wrong name in the React debug extensions for Chrome/Firefox.
6. Similarly for map controllers, use an accurate class name, e.g., `export default class DualMapController ...` (not `... MapController ...`).

I know the existing situation is largely legacy, but it's starting to get confusing and ever more inconsistent and this is the time to fix it.

I experimented on a separate branch with renaming, etc., and github is smart enough to follow the renames logically. Yay. Otherwise this would have to go in a separate PR. Just in case something does screw up and the renaming blows up on us, please do all this renaming in separate commit(s) so that they can be backed out easily, and can be cherry-picked over to a separate branch if necessary.

rod-glover:
@corviday , please note that there are comments in the Files Changed view that for some reason don't appear here. In particular, [this one](https://github.com/pacificclimate/climate-explorer-frontend/pull/141/files#r172705537) is important.

src/components/MapController/DualMapController/DualMapController.js
```diff
@@ -61,21 +80,10 @@ export default class MapController extends React.Component {
...
-  }
-
-  hasComparand() {
-    return this.hasValidData('comparand');
-  }
 
   // TODO: https://github.com/pacificclimate/climate-explorer-frontend/issues/118
-  currentDataset() {
-    // Return encoding of currently selected dataset
+  currentInstance() {
```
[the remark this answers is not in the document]
jameshiebert: Yeah, I've never really liked the use of "instance". Instance usually refers to *one* thing, so using it to describe a set of selecting filters that could describe multiple sets of data doesn't really match.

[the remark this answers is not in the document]
corviday: I think datasetSpec would be confusing, since in the context of of ce frontend, "dataset" means "a single file with data in it" and this specification doesn't uniquely describe a single file. It uniquely describes a single - I don't know what to call it, which is sort of how it got called instance - chunk of model output, either a single run or a single run and climatology period, but there can be multiple files representing different views of that output and the specification describes all of them.

How do you feel about just dataSpec?

Alternately, we could call single files datafiles and "set of files representing a specific data chunk and / or the parameters that uniquely describe that chunk" datasets, they'd be a set of files.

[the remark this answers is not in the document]
rod-glover: dataSpec is just fine with me.

src/components/AppController/AppController.js
rod-glover: @corviday , I would be happy to do this task for you. My IDE has refactoring tools that do 99% of the bookkeeping work associated with this (modifying all the references to the renamed files, etc.), which I know is a real pain to do manually.

src/components/AppMixin/AppMixin.js
```diff
@@ -59,6 +59,14 @@ var AppMixin = {
             var start = timestampToYear(response.data[key].start_date);
             var end = timestampToYear(response.data[key].end_date);
 
+            // Stopgap measure to deal with the fact that experiment string formats
+            // vary between climdex files ("historical, rcp26") and GCM outputs
+            // ("historical,rcp26"). Formats experiment strings to include a space.
+            // This formatting is undone to run queries against the database by
+            // ce-backend.guessExperimentFormatFromVariable()
+            // TODO: remove this when no longer needed.
+            var normalizedExp = String(response.data[key].experiment).replace(',r', ', r');
```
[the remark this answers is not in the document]
corviday: A more generic or flexible substitution here does not give us any more functionality, because later on we have to *exactly* undo this substitution in order to run queries against the database with whatever the original database string was, which is of course the one in the database. If there were multiple possible pre-standardization scenario strings (zero spaces, three spaces, underscores, whatever) for each standard one, we would not necessarily be able to recover the original string when we needed it for the back end.

If at some point we have more than two possible spellings of each emissions scenario description, we'll need a dictionary or something else more complex.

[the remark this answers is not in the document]
rod-glover: Alternatively, and in my opinion better, the backend needs to be smarter about these strings, performing a similar normalization as part of the matching. This should not be hard. And it follows the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle), "Be conservative in what you send, be liberal in what you accept".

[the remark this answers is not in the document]
corviday: We have a [data prep issue](https://github.com/pacificclimate/climate-explorer-data-prep/issues/48) about experiment strings, should I add a backend issue too / instead?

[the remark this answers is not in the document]
rod-glover: Yes, please.

src/components/MapController/DualMapController/DualMapController.js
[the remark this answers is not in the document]
jameshiebert: dataSpec works for me.

src/components/AppMixin/AppMixin.js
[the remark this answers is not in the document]
corviday: Here it is: https://github.com/pacificclimate/climate-explorer-backend/issues/76

corviday:
- [ ] Replace == with === across changes made
- [ ] Add screenreader ids for accessibility to components React is complaining about (not that an app focussed on maps and graphs is going to make any sense with a screen reader)
- [ ] Fix handleChange proptype warning
- [ ] Rewrite layer parameters function change
- [ ] replace "instance" with "dataSpec"
- [ ] pass top level selection parameters to MapControllers
- [ ] Rename and reorganize MapControllers, DataControllers, and AppControllers
- [ ] More MapController DRYing
- [ ] Remove comments left from previous refactoring
- [ ] Use const as appropriate across changes made

src/components/MapController/DualMapController/DualMapController.js
```diff
@@ -289,45 +251,25 @@ export default class MapController extends React.Component {
     // the first dataset representing a 0th time index (January, Winter, or Annual)
     // will be displayed.
 
-    if (this.hasValidData('variable', nextProps)) {
+    if (hasValidData('variable', nextProps)) {
       // TODO: DRY this up
-      const newVariableId = nextProps.meta[0].variable_id;
-      const oldVariableId = this.props.meta.length > 0 ? this.props.meta[0].variable_id : undefined;
-      const hasComparand = nextProps.comparandMeta && nextProps.comparandMeta.length > 0;
+      const newVariableId = selectedVariable(nextProps.meta);
```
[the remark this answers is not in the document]
corviday: >  4. Any reason this is not a good idea?

Unlike `variable_id`, MapControllers and their children never determine or use `experiment` or `model_id`. It wouldn't hurt anything to pass them in as props, but I don't think there's any particular reason to at the moment.

rod-glover:
Just to repeat my offer: I'd be happy to rename and reorganize MapControllers, DataControllers, and AppControllers

corviday:
[Here is the new version, for testing.](http://docker2.pcic.uvic.ca:30302/#/precipitation/extreme_precipitation)

src/components/DataSpecSelector/DataSpecSelector.js
```diff
@@ -0,0 +1,61 @@
...
+    return ids;
+  }
+
+  render() {
+    const dataSpecs = this.makeDataSpecs();
+    return (
+      dataSpecs.length > 0 &&
+      <Selector
+        label='Select Dataset'
+        items={dataSpecs}
```
rod-glover: Maybe just inline `dataSpecs`?

src/components/MapSettings/TimeSelector.js
```diff
@@ -41,7 +41,7 @@ export default class TimeSelector extends React.Component {
     );
   }
 
-  static timeLinkedTooltip = <Tooltip>Timestamp matching is activated</Tooltip>;
+  static timeLinkedTooltip = <Tooltip id={"Time Link Button"}>Timestamp matching is activated</Tooltip>;
```
rod-glover: HTML id attr [must not contain whitespace](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)
Also, for pure string props, you don't have to use { } and typically people don't.

src/components/data-controllers/SingleDataController/SingleDataController.js
```diff
@@ -176,7 +174,7 @@ var DataController = createReactClass({
         {
           multiYearMeanSelected(this.props) ? (
 
-            <Tabs>
+            <Tabs id={"Graphs"}>
```
rod-glover: Thanks for fixing all those missing required id attributes.

src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
```diff
@@ -58,17 +58,22 @@ export default class AnnualCycleGraph extends React.Component {
...
+      dataSpec: { start_date, end_date, ensemble_member },
       graphSpec: blankGraphSpec,
     };
   }
 
-  firstMonthlyMetadata({ meta, model_id, variable_id, experiment }) {
-    return _.findWhere(
-      meta,
-      { model_id, variable_id, experiment, timescale: 'monthly' }
+  initialDataSpec({ meta, model_id, variable_id, experiment }) {
```
rod-glover: nice

src/components/graphs/DualAnnualCycleGraph.js
```diff
@@ -22,49 +25,43 @@ export default function DualAnnualCycleGraph(props) {
...
-      monthlyVariableMetadata, { timescale: 'seasonal' }, meta
-    );
-    const yearlyVariableMetadata = findMatchingMetadata(
-      monthlyVariableMetadata, { timescale: 'yearly' }, meta
-    );
-
-    let metadataSets = [
+    const monthlyVariableMetadata = findMetadataForDataSpec(variable_id, 'monthly', meta);
+    const seasonalVariableMetadata = findMetadataForDataSpec(variable_id, 'seasonal', meta);
+    const yearlyVariableMetadata = findMetadataForDataSpec(variable_id, 'yearly', meta);
```
rod-glover: :+1:

src/components/graphs/DualAnnualCycleGraph.js
```diff
@@ -22,49 +25,43 @@ export default function DualAnnualCycleGraph(props) {
...
+      if(seasonalVariableMetadata) {
+       metadataSets = metadataSets.concat(
+           findMetadataForDataSpec(comparand_id, 'seasonal', comparandMeta));
+      }
+      if(yearlyVariableMetadata) {
+        metadataSets = metadataSets.concat(
+            findMetadataForDataSpec(comparand_id, 'yearly', comparandMeta));
+      }
+    }    
+    return _.compact(metadataSets);
```
rod-glover: elegant

src/components/map-controllers/map-helpers.js
```diff
@@ -0,0 +1,191 @@
...
+  params.palette = 'x-Occam';
+  return params;
+}
+
+/**************************************************************************
+ * 2. State handling functions shared by MapControllers
+ **************************************************************************/
+
+// TODO: https://github.com/pacificclimate/climate-explorer-frontend/issues/118
+function currentDataSpec() {
```
rod-glover: This function is defined twice in this file.

src/components/map-controllers/PrecipMapController/PrecipMapController.js
```diff
@@ -0,0 +1,293 @@
...
+  // Handlers for dataSpec change
+
+  // TODO: https://github.com/pacificclimate/climate-explorer-frontend/issues/118
+  updateDataSpec = (encodedDataSpec) => {
+    this.loadMap(this.props, JSON.parse(encodedDataSpec));
+  };
+
+  // Handlers for settings changes
+
+  handleChangeTime = this.updateTime.bind(this);
```
rod-glover: Simplify: Declare `updateTime` as an arrow function and rename it `handleChangeTime`.
Anywhere you are binding only to `this`, just use an arrow function instead, which does the binding for you.

src/components/DataMap/DataMap.js
```diff
@@ -134,25 +137,33 @@ class DataMap extends React.Component {
...
+      const handlerName = `handle${layertype.charAt(0).toUpperCase() + layertype.slice(1)}LayerRef`;
+      return (
+          <DataLayer
+            layerType={layertype}
+            {...this.props[layertype]}
+            onLayerRef={this[handlerName]}
+          />
+      );
+    }
+  }
```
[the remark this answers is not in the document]
rod-glover: Alternatively, leave `DataLayer` as it is and create a [stateless functional component](https://reactjs.org/docs/components-and-props.html#functional-and-class-components) to use inside the `map()`:

```
const WrapDataLayer = props => {
   const { layerType, layerParams, onLayerRef } = props;
   if (!layerParams) {
      return null;
   }
   return <DataLayer  layerType=layerType, ...layerParams, onLayerRef=onLayerRef>;
}
```

I don't prefer this at all in this case, but if refactoring `DataLayer` is a problem for some reason, then its an alternative. Classic React component wrapping another component.

In general, when you are tempted to define a method on a component that builds another component, turn that method into a separate component. They are often stateless functional components, which are simple and which React renders particularly fast (no lifecycle).

corviday:
I think I hit all the suggested changes.

rod-glover:
Code changes look good.

Regrettably, tests are failing. 

I suspect DataLayer-smoke is just failing on incorrect test setup.

There are also a lot of Failed prop type messages, which may or may not cause Travis to fail the build, but which do indicate problem(s) that ought to be addressed. Suspect it's pretty simple.

corviday:
Aaaaaaargh. I don't know how I forgot that.

rod-glover:
Happens to me about 3 times a day. No sweat.

rod-glover:
LGTM. Good work.

As we discussed, there are still a few lingering questions about prop types, but those are best left for a more general review around that subject.

jameshiebert:
Excellent! I think that this wraps up all of our work on climate-explorer-frontend for FY17-18. @rod-glover , could you merge this, roll a new tagged release (with release notes), wait for docker hub to build the tag and then deploy it to docker-dev01 (replacing the current `cef-beta` image)?

corviday:
The live ncwms doesn't serve its files yet.

jameshiebert:
> The live ncWMS doesn't serve its files yet.

OK, want to take that on. You guys can work in parallel.

jameshiebert:
>  @rod-glover , could you merge this, roll a new tagged release (with release notes), wait for docker hub to build the tag and then deploy it to docker-dev01 (replacing the current cef-beta image)

I'll take this on to get it done ASAP.

Issue #145, opened 2018-04-11
Add present / future comparison plot


corviday:
Add a plot to Single Data Controller comparing present (2010-2039 climatology) and future annual cycles.

corviday:
![future](https://user-images.githubusercontent.com/4512605/38642905-6287c6ce-3d90-11e8-9345-03eabf99c417.png)

I've done this the obvious way: just putting multiple annual cycle datasets on the same graph, but the scale of the variation between summer and winter is so much larger then the variation between June 2010 and June 2070 that the latter is completely swamped, visually. I don't know how useful it would be to a viewer. 

I need to think a bit about whether there's a better way to do this. Possibly something like a Pareto graph, showing the delta between climatology as a separate time series above or below the main annual cycle?

Issue #70, opened 2017-09-26
Test and refine DataController layout on small screens


corviday:
The Download Data controls and the Select Time of Year controls seemed to render on top of each other on a low-resolution laptop screen. Should look into that and if necessary adjust the [bootstrap Grid](https://v4-alpha.getbootstrap.com/layout/grid/) parameters (separate layouts for small, medium, and large screens)  to make that work better.

corviday:
Matthew says:

> I just noticed that the map on the current climate explorer page doesn't seem to be autoscaling - it's fixed. So when viewing on smaller screens it floats over top of the other content.
> 
> It looks like the map has been fixed at 890 x 700px

rod-glover:
Fixed

PR #146, opened 2018-04-17
Add a future annual cycle graph

corviday:
[Adds a present / future comparison graph to single-variable portals showing future values.](https://github.com/pacificclimate/climate-explorer-frontend/issues/145) The difference between January 2010 and January 2040 is _much_ smaller than the difference between January 2010 and June 2010 and gets lost in the noise. So in addition to the usual annual cycle time series for each climatology, there is a row of bar charts along the bottom, showing the difference between each climatology and the present-day base climatology. This lets a user see, in more detail, "Oh, it will be warmer in January but colder in March according to this model."

screenshot from 2018-04-17 14-05-44

This PR is much smaller than it looks. Most of the putatively new or deleted lines actually stem from splitting the over-long `chart.js` into three files, `chart-generators.js`, `chart-formatters.js`, and `chart-transformers.js`, which was straightforward and didn't require rewriting anything beyond file paths. 

Actual new code is the following:

- `chart-formatters.js`: `padYAxis()` and `hideSeriesInTooltip()`
- `chart-transformers.js`: `makeAnomalyGraph()` and `addAnomalyTooltipFormatter()`
- `AnomalyAnnualCycleGraph.js`
- `SingleDataController.js` : the section in render() that calls AnomalyAnnualCycleGraph
- `chart-formatter-tests.js` : tests for `padYAxis()` and `hideSeriesInTooltip()`
- `chart-transformer-tests.js`: tests for `makeAnomalyGraph()` and `addAnomalyTooltipFormatter()`

Everything else is from splitting up chart.js, splitting up chart.js's tests, or editing imports to point to where chart functions ended up.

### DIFF
--- a/src/core/chart-formatters.js
+++ b/src/core/chart-formatters.js
@@ -184,7 +184,6 @@ var sortSeriesByRank = function(graph, ranker) {
   return graph;
 };
 
-//TODO: this function needs a test.
 /*
  * Post-processing graph function that hides specific series from the tooltip.
  * 
@@ -220,7 +219,6 @@ var hideSeriesInTooltip = function(graph, predicate) {
   return graph;
 }
 
-//TODO: this function needs a test
 /*
  * Post-processing graph function that adds extra space above or below
  * data on a graph by setting the y-axis maximums and minimums to multiples
@@ -236,11 +234,15 @@ var hideSeriesInTooltip = function(graph, predicate) {
  */
 var padYAxis = function (graph, axis = "y", direction = "top", padding = 1) {
   if(padding <= 0) {
-    throw new Error("Graph axis padding value must be greater than 0");
+    throw new Error("Error: Graph axis padding value must be greater than 0");
   }
   
   if(direction != "top" && direction != "bottom") {
-    throw new Error("Unknown graph axis padding direction");
+    throw new Error("Error: Unknown graph axis padding direction");
+  }
+  
+  if(axis !== "y" && axis !== "y2") {
+    throw new Error("Error: invalid scaling axis");
   }
   
   // if this graph does not yet have minimums and maximums defined, calculate
--- a/src/core/chart-formatters.js
+++ b/src/core/chart-formatters.js
@@ -231,6 +231,16 @@ function hideSeriesInTooltip (graph, predicate) {
 /****************************************************************************
  * 1. Axis formatters
  ****************************************************************************/
+/*
+ * Helper function that returns an array of all data series associated with 
+ * a specific y axis (y or y2). Ignores category or time series, if present.
+ */
+function getDataSeriesByAxis(graph, axis) {
+  return _.filter(graph.data.columns, series => {
+    const seriesName = series[0];
+    return seriesName !== 'x' && graph.data.axes[seriesName] === axis;
+  });
+}
 
 /*
  * Post-processing graph function that adds extra space above or below
@@ -262,9 +272,7 @@ function padYAxis (graph, axis = "y", direction = "top", padding = 1) {
   // them from the data.
   let min = graph.axis[axis].min;
   let max = graph.axis[axis].max
-  const axisSeries = _.filter(graph.data.columns, series => {
-      return series[0] !== 'x' && graph.data.axes[series[0]] === axis
-    });
+  const axisSeries = getDataSeriesByAxis(graph, axis);
   
   if(_.isUndefined(min)) {
     min = _.min(_.map(axisSeries, series => _.min(series)));
@@ -296,9 +304,7 @@ function padYAxis (graph, axis = "y", direction = "top", padding = 1) {
 
 function hideTicksByRange(graph, axis = "y", min, max) {
   const oldFormatFunction = graph.axis[axis].tick.format;
-  const axisSeries = _.filter(graph.data.columns, series => {
-    return series[0] !== 'x' && graph.data.axes[series[0]] === axis
-  });
+  const axisSeries = getDataSeriesByAxis(graph, axis);
   
   //if a range is not supplied, generate one from the data
   const genMin = _.isUndefined(min);
@@ -325,4 +331,7 @@ function hideTicksByRange(graph, axis = "y", min, max) {
 
 module.exports = { assignColoursByGroup, fadeSeriesByRank,
     hideSeriesInLegend, sortSeriesByRank, hideSeriesInTooltip,
-    padYAxis, hideTicksByRange};
+    padYAxis, hideTicksByRange,
+    //helper functions exported only for testing:
+    getDataSeriesByAxis
+    };
--- a/src/core/chart-formatters.js
+++ b/src/core/chart-formatters.js
@@ -3,12 +3,12 @@
  *   to alter the way data is displayed to make charts more readable. These
  *   functions do not affect the data itself, only its formatting and display.
  * 
- * Each function in this file accepts a C3 graph specification object and a
+ * Data series format functions accept a C3 graph specification and a 
  * segmentation function. The segmentation function will be applied to each
  * data series in the C3 graph object, with the results being used to decide
- * how to format data from that series. It returns the modified graph spec.
+ * how to format data from that series. 
  * 
- * The functions in this file are:
+ * Data series formatters:
  *  - assignColoursByGroup: assigns the same display colour to all data series
  *      belonging to the same group
  *      
@@ -21,8 +21,15 @@
  *      lower ranked series
  *      
  *  - hideSeriesInTooltip: removes specific data series from the tooltip
- *      
+ * 
+ * Axis formatters accept a C3 graph specification and additional parameters
+ * that vary by function. They adjust formatting on graph axes.
+ * 
+ * Axis formatters:
  *  - padYAxis: add additional blank space above or below the data series
+ *  
+ *  - displayTicksByRange: only display axis ticks for specific parts of the
+ *    data range
  ***************************************************************************/
 import _ from 'underscore';
 import {PRECISION,
@@ -33,6 +40,10 @@ import {PRECISION,
         getVariableOptions} from './util';
 import chroma from 'chroma-js';
 
+
+/****************************************************************************
+ * 0. Data series formatters
+ ****************************************************************************/
 /*
  * Reiteration of D3's "category10" colors. They underlie c3's default
  * colours but are not directly accessible. Allows creating custom
@@ -217,6 +228,10 @@ function hideSeriesInTooltip (graph, predicate) {
   return graph;
 }
 
+/****************************************************************************
+ * 1. Axis formatters
+ ****************************************************************************/
+
 /*
  * Post-processing graph function that adds extra space above or below
  * data on a graph by setting the y-axis maximums and minimums to multiples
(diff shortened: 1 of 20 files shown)


rod-glover:
A few questions that might have been resolved already in discussions I wasn't part of:

1. Why bar charts (as opposed to line graphs)?
2. How many climatologies might be shown on this graph, and thus in the difference display at the bottom? If more than 2 or 3, this could get awkward for bar charts. Possibly, but not certainly, a set of line graphs might be easier to read in those cases.
3. Is the vertical space allocated for the difference display large enough? It seems like it might be hard to read the differences very accurately.

src/core/chart-formatters.js
```diff
@@ -0,0 +1,284 @@
...
+ * axis, but don't want them to visually overlap.
+ * 
+ * Arguments:
+ *   graph - the graph to be modified
+ *   axis - either "y1" or "y2"
+ *   direction - where to add padding, either "top" or "bottom"
+ *   padding - the amount of extra y-axis space to add, expressed as a 
+ *             multiple of the existing data span. 
+ */
+var padYAxis = function (graph, axis = "y", direction = "top", padding = 1) {
```
rod-glover: Not sure where this habit of declaring functions as variables got started, but it's unnecessary and in some environments can result in anonymous functions where you really want a named function. (I think Babel is smart about naming functions, but some environments aren't and these kinds of anonymous functions make debugging a lot harder.) 

Much prefer a named function, i.e.,
```
function padYAxis(...) { ... }
```

rod-glover: On further reflection, I'm not sure if fixing this issue throughout belongs in this PR. It's always tempting to slip in some code improvement like this, but it does add unnecessary changes.

src/core/chart-formatters.js
```diff
@@ -0,0 +1,284 @@
...
+ */
+var hideSeriesInTooltip = function(graph, predicate) {
+  //determine which series do not appear in the tooltip
+  let hidden = [];
+  
+  for(let i = 0; i < graph.data.columns.length; i++) {
+    if(predicate(graph.data.columns[i])) {
+      hidden.push(graph.data.columns[i][0]);
+    }
+  }
```
rod-glover: This could be simplified as
```
const hidden = _.pluck(_.filter(graph.data.columns, predicate), 0);
```

src/core/chart-formatters.js
```diff
@@ -0,0 +1,284 @@
...
+  let max = graph.axis[axis].max
+  if(_.isUndefined(min)) {
+    min = Infinity;
+    for(let i = 0; i < graph.data.columns.length; i++) {
+      let series = graph.data.columns[i];
+      if(series[0] !== 'x' && graph.data.axes[series[0]] === axis) {
+        const seriesMin = _.min(series);
+        min = seriesMin < min ? seriesMin : min; 
+      }
+    }
```
rod-glover: This loop could be eliminated and simplified as:
```
min = _.min(
   _.map(
      _.filter(graph.data.columns, series => series[0] !== 'x' && graph.data.axes[series[0]] === axis),
      series => _.min(series)
   )
);
```

Better with chaining:
```
_.chain(graph.data.columns)
.filter(graph.data.columns, series => series[0] !== 'x' && graph.data.axes[series[0]] === axis)
.map(_.min)
.min()
.value();
```

Alternatively:
```
_.chain(graph.data.columns)
.filter(graph.data.columns, series => series[0] !== 'x' && graph.data.axes[series[0]] === axis)
.flatten()
.min()
.value();
```

src/core/chart-formatters.js
```diff
@@ -0,0 +1,284 @@
...
+  
+  if(_.isUndefined(max)) {
+    max = -Infinity;
+    for(let i = 0; i < graph.data.columns.length; i++) {
+      let series = graph.data.columns[i];
+      if(series[0] !== 'x' && graph.data.axes[series[0]] === axis) {
+        const seriesMax = _.max(series);
+        max = seriesMax > max ? seriesMax : max; 
+      }
+    }
```
rod-glover: Similarly

corviday:
[Here's a docker for this branch](http://docker-dev01.pcic.uvic.ca:30080/#/). 

I'll take all the UI feedback I can get. :)

src/core/chart-formatters.js
rod-glover: And you can factor out the `series` predicate that is common to both computations. That would make the `underscore` expressions tighter and probably possible to put on one line with clarity.

src/core/chart-transformers.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+  graph.axis.y2.label = {};
+  graph.axis.y2.label.position = 'outer-middle';
+  graph.axis.y2.label.text = `${getAxisTextForVariable(graph, baseSeries[0])} anomaly from ${baseSeries[0]}`;
+  
+  
+  // function that determines whether a data series an anomaly series.
+  // used to format anomalies differently than data
+  let isAnomaly = function (series) {
```
rod-glover: Please use just a function declaration: `function isAnomaly(series) { ... }`.

Also, `const` here would be the appropriate declaration; `let` is for values that will be reassigned, i.e., that are variable and that does not seem to be your intention here.

src/core/chart-transformers.js
```diff
@@ -0,0 +1,283 @@
...
+    return series[0].indexOf("Anomaly") != -1;
+  }
+  
+  // classifier function that matches each "anomaly" data series with
+  // the nominal series it is based on. Used to match colours.
+  let anomalyMatcher = function(series) {
```
rod-glover: Ditto `function`

src/core/chart-transformers.js
```diff
@@ -0,0 +1,283 @@
...
+    let sName = series[0];
```
rod-glover: `const`

src/core/chart-transformers.js
```diff
@@ -0,0 +1,283 @@
...
+    return isAnomaly(series) ? sName.substring(0, sName.length - 8) : sName;
+  };  
+  
+  // ranking function that assigns anomaly series lower results than 
+  // nominal series, used to make them distinguishable.
+  let anomalyRanker = function (series) {
```
rod-glover: Ditto `function`

rod-glover:
More thoughts on UI:

1. In the difference display, I'd eliminate the means, which by definition always have difference value zero.
2. I'd definitely allocate more space to the difference display; at least twice what is there now, possibly much more. (I think, but am not sure, that this display is roughly equally important to the absolute value display above it. Thoughts?)
3. I think I still prefer the line graph for differences, but let's compare the options with a larger area for the difference display.

src/core/chart-formatters.js
corviday: Ooo, nice.

corviday:
I've redone [the docker](http://docker-dev01.pcic.uvic.ca:30080/#/) for this branch and changed the graph formatting to make the difference graph and the absolute value about the same size.

As a way of making it more obvious which axis goes with which graph, I also experimented with blanking out some of the axis labels.

Opinions?

src/core/chart-formatters.js
```diff
@@ -0,0 +1,328 @@
...
+  // them from the data.
+  let min = graph.axis[axis].min;
+  let max = graph.axis[axis].max
+  const axisSeries = _.filter(graph.data.columns, series => {
+      return series[0] !== 'x' && graph.data.axes[series[0]] === axis
+    });
+  
+  if(_.isUndefined(min)) {
+    min = _.min(_.map(axisSeries, series => _.min(series)));
+  }
```
rod-glover: 👍

src/core/chart-formatters.js
```diff
@@ -0,0 +1,328 @@
...
+    throw new Error("Error: invalid scaling axis");
+  }
+  
+  // if this graph does not yet have minimums and maximums defined, calculate
+  // them from the data.
+  let min = graph.axis[axis].min;
+  let max = graph.axis[axis].max
+  const axisSeries = _.filter(graph.data.columns, series => {
+      return series[0] !== 'x' && graph.data.axes[series[0]] === axis
+    });
```
rod-glover: Since this code is repeated verbatim at least one place elsewhere, I'd pull it out into a common function. Possibly that function would take the axis identifier ('x') as a parameter.

src/core/chart-transformers.js
```diff
@@ -0,0 +1,287 @@
...
+ */
+function makeVariableResponseGraph (x, y, graph) {
+  let c3Data = {};
+
+  const seriesNameContains = function (series, keyword) {
+    return caseInsensitiveStringSearch(series[0], keyword);
+  }
+  
+  const xseries = _.filter(graph.data.columns, series => seriesNameContains(series, x));
+  const yseries = _.filter(graph.data.columns, series => seriesNameContains(series, y));
```
rod-glover: 👍

src/core/chart-transformers.js
```diff
@@ -0,0 +1,287 @@
...
+    c3Data.columns[1].push(tuples[i][1]);
+    //C3 doesn't really support scatterplots, but we can fake it by adding
+    //a missing data point between each actual data point, and instructing C3
+    //not to connect across missing data points with {connectNull: false} 
+    //TODO: breake this out into makeScatterplot(); we'll likely need it again
+    if(i < tuples.length - 1) {
+      c3Data.columns[0].push((tuples[i][0] + tuples[i+1][0])/2);
+      c3Data.columns[1].push(null);
+    }
+  }
```
rod-glover: Wherever you find yourself iterating through an array and pushing items onto another array derived from the iterated array, you can usually reformulate it more clearly as a set of filter, map, and other operations (courtesy of underscore). 

When the set of filter and map operations gets larger that 2 or 3 (or even 1!), you can use chaining (part of underscore) to make it clearer: Chaining means that the array arguments you'd have to pass explicitly to the filter, map, etc. functions are omitted because the operations are methods of the array (wrapped as a chainable object by underscore). It's very nice, and close to the syntax of some of the more delightful functional languages.

rod-glover:
Lee, nice work. 

UI thoughts:

1. The larger space allocated for the difference display is much better to my eye. Perhaps the team should decide what fraction; I kinda like the size it is now, but it could be smaller and still have the same impact I think.
2. I like the blanked axis labels, they make things clearer. As an alternative, and only just thinking aloud here, could there be two different graphs, with the x-axis of one somehow omitted or suppressed. That might simplify the label juggling if it's possible.
3. In the difference display, can the baseline values (with constant difference zero) be removed?
4. I noticed that on my laptop screen, the layout doesn't work very well: the map overlaps the graphs. Probably just an easy Bootstrap fix.

corviday:
2. I did them as one graph so they'd share the tooltip, which gives both the absolute and difference values, and provides a line connecting the absolute and difference graphs. I think the tooltip is pretty great, but I'm open to doing two separate graphs if we prefer.

3. Yes, definitely. I personally found the difference graph is easier to understand with the zero-difference line providing an ersatz axis, but at this point, I've reworked this graph enough I'm not sure my intuitions about it are useful. :) 

4. This is issue https://github.com/pacificclimate/climate-explorer-frontend/issues/70 

I'll tidy the code and get up a version without the zero-difference line Monday.

rod-glover:
2. Ah, the shared tooltip is a good idea. Probably worth retaining.
3. I get the value of the zero axis. Could it be achieved with just two points and one line segment joining them?

corviday:
[Docker for this branch](http://docker-dev01.pcic.uvic.ca:30080/#/) has been updated and is now caught up to all the UI suggestions! Removed the "anomaly" line for the 2010 climatology (all zeroes) and switched the bar charts to line charts.

src/core/chart-formatters.js
```diff
@@ -0,0 +1,328 @@
...
+function hideSeriesInLegend (graph, predicate) {
+  var hiddenSeries = [];
+
+  _.each(graph.data.columns, column => {
+    var seriesName = column[0];
+    if(!_.isEqual(seriesName, "x")) {
+      if(predicate(column)){
+        hiddenSeries.push(seriesName);
+      }
+    }
```
rod-glover: Ditto comments re. `_.each` vs. for-of, and `!==` vs. `isEqual` for simple values.
Also, the nested if statements should be collapsed to one.

FWIW, this could be written as
```
const hiddenSeries = 
   _.chain(graph.data.columns)
   .filter(column => column[0] !== 'x' && predicate(column))
   .map(column => column[0]);
```
which depending on reader might be deemed easier to read.

src/core/chart-formatters.js
rod-glover: No?

src/core/chart-transformers.js
rod-glover: I see your point; I didn't quite notice the double pushes. Just for reference, the double pushes could be accommodated by mapping to pairs (two-element arrays) and then flattening the result. (In this case, you'd also need to push the final two results afterwards. That plus other minor complications makes my functional solution sound worse than your iterative solution in this case.)

I don't see your reduce solution here, but it doesn't _sound_ like a better solution.

rod-glover:
I like the new Future Anomaly graph better.

The y-ranges for value and difference graphs overlap; can they be made entirely disjoint? The overlap is confusing, especially when (as I see in the default model, var, emission), the first difference actually extends into the range of the value graph (looks like -20 C). It would be better to have some blank space between the two graphs.

src/core/chart-transformers.js
corviday: Sorry about that; I thought github would update with what I'd changed this section to. [Here's the reduce solution.](https://github.com/pacificclimate/climate-explorer-frontend/pull/146/files#diff-dc15f875105ec95e990b237fac511880R94) I wasn't terribly happy with it.

rod-glover:
Nice work, Lee.

PR #148, opened 2018-04-18
Pass only needed data to context graphs

corviday:
Previously, `SingleAppController` passed metadata for _all_ datasets in the current ensemble to `SingleDataController` as `props.contextMeta`. The metadata collection was used to select datasets to display in the Model Context graph comparing results from all models for the selected scenario and variable. We don't actually need or want most of that. Filter by scenario and variable.

### DIFF
--- a/src/components/app-controllers/SingleAppController/SingleAppController.js
+++ b/src/components/app-controllers/SingleAppController/SingleAppController.js
@@ -12,6 +12,7 @@
 import React from 'react';
 import createReactClass from 'create-react-class';
 import { Grid, Row, Col } from 'react-bootstrap';
+import _ from 'underscore';
 
 import SingleMapController from '../../map-controllers/SingleMapController';
 import SingleDataController from '../../data-controllers/SingleDataController/SingleDataController';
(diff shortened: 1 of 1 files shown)


src/components/app-controllers/SingleAppController/SingleAppController.js
```diff
@@ -36,6 +37,13 @@ export default createReactClass({
     return !(datafile.multi_year_mean == false && datafile.timescale == "monthly");
   },
 
+  //Returns metadata for datasets with thethe selected variable + scenario, any model.
+  //Passed as a prop for SingleDataController to generate model comparison graphs.
+  getModelContextMetadata: function () {
+    return _.where(this.state.meta,
+        {variable_id: this.state.variable_id, experiment: this.state.experiment});
+  },
+
```
rod-glover: This is good. I'm not sure whether to prefer the ES6 native [Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) over a more conveniently formulated underscore function. Your call.

PR #149, opened 2018-04-19
Add stick plot

corviday:
This PR builds on (and includes) some of the changes to graph generation made in the [Add A Future Annual Cycle Graph PR](https://github.com/pacificclimate/climate-explorer-frontend/pull/146), so that one should be merged first. It benefits from, but is not strictly dependent on, the [Pass Only Needed Data To Context Graphs PR](https://github.com/pacificclimate/climate-explorer-frontend/pull/148).

Creates a vertical "stick plot" showing the values output by various models for identical inputs (variable, emissions scenario, date) as a way give a user a sense of where in the consensus the particular model they are viewing falls.

It's currently displaying as a stand-alone graph, which almost certainly isn't what we'll want to do with it long term. We'll want to use it as a sidebar to provide context to some viewer displaying data from a single model.
sidebar

### DIFF
--- a/src/components/data-controllers/SingleDataController/SingleDataController.js
+++ b/src/components/data-controllers/SingleDataController/SingleDataController.js
@@ -47,6 +47,7 @@ import SingleContextGraph from '../../graphs/SingleContextGraph';
 import SingleTimeSeriesGraph from '../../graphs/SingleTimeSeriesGraph';
 import { getStats } from '../../../data-services/ce-backend';
 import AnomalyAnnualCycleGraph from '../../graphs/AnomalyAnnualCycleGraph';
+import SingleTimeSliceGraph from '../../graphs/SingleTimeSliceGraph';
 
 // TODO: Remove DataControllerMixin and convert to class extension style when 
 // no more dependencies on DataControllerMixin remain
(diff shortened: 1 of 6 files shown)


corviday:
I experimented with using the C3 API to `select()` the point representing the model in question, instead of displaying it as a different colour. `Select()`ing a point draws a circle around it, and I felt that the graph was a little easier to read that way.

Ultimately, however, this was against the philosophy C3 built into their chart API. `Select()` is intended to facilitate human interaction, so using it for passive display required a lot of overloading default selection functionality, and this led to some very roundabout code. It seemed like circling a point instead of colouring it wasn't enough gain in exchange for some strange and probably fragile code, so I went back to indicating a point by display colour.

But if users report a hard time reading the graph, there are alternatives.

rod-glover:
This shares commits with #146. Please rebase either onto that PR's branch or onto master after the PR is merged.

corviday:
[Docker for this branch](http://docker-dev01.pcic.uvic.ca:30080/#/). The new graph is on the tab marked "Snapshot".

jameshiebert:
This all looks straightforward. Let's merge it!

PR #151, opened 2018-04-23
Use correct scenario string for downscaled data queries to multistat

corviday:
Downscaled datasets typically have emissions scenario strings of the form "`historical,rcp26`" as opposed to the "`historical, rcp26`" format used by climdex datasets. Climate explorer frontend was using the spaced format for all datafiles when querying the `multistat` API endpoint. This PR switches to the unspaced variant for downscaled datasets.

This is a short-term fix. Long term we'd like to [standardize emission strong formats in the backend](https://github.com/pacificclimate/climate-explorer-backend/issues/76).

### DIFF
--- a/src/components/DataControllerMixin/DataControllerMixin.js
+++ b/src/components/DataControllerMixin/DataControllerMixin.js
@@ -8,7 +8,6 @@
  * provided by this mixin.
  * 
  * Provides functions to:
- * - fetch data from the backend API
  * - initialize data state on component loading
  * - export data to csv or xls file
  * - filter and manipulate data
@@ -16,7 +15,6 @@
 
 import _ from 'underscore';
 import urljoin from 'url-join';
-import axios from 'axios';
 import {exportDataToWorksheet, 
         generateDataCellsFromC3Graph} from '../../core/export';
 import {validateLongTermAverageData,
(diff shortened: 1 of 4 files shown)


src/components/DataControllerMixin/DataControllerMixin.js
```diff
@@ -74,56 +74,39 @@ var ModalMixin = {
...
-      var filtered = {};
+  /*
+   * Filters data from any call to the API that returns an object with
+   * individual values keyed to unique_ids. It returns a new object
+   * that contains only results from datasets whose metadata matches the
+   * attributes passed to the filter argument.
+   * 
+   * This is a temporary stopgap until this issue is solved:
+   * https://github.com/pacificclimate/climate-explorer-backend/issues/61
+   * FIXME: remove this function and calls to it when issue 61 is solved.
```
rod-glover: https://github.com/pacificclimate/climate-explorer-backend/issues/61 is marked as closed by https://github.com/pacificclimate/climate-explorer-backend/pull/67, so this function should be removed as per FIXME.

src/components/data-controllers/MotiDataController/MotiDataController.js
```diff
@@ -108,7 +110,7 @@ export default createReactClass({
     
     //Load stats table
     this.setStatsTableNoDataMessage("Loading Data");
-    var myStatsPromise = this.getStatsPromise(props, this.state.dataTableTimeOfYear);
+    var myStatsPromise = getStats(props, this.state.dataTableTimeOfYear).then(validateStatsData);
```
rod-glover: 👍

src/data-services/ce-backend.js
```diff
@@ -59,6 +59,42 @@ function getData(
   });
 }
 
+// TODO: https://github.com/pacificclimate/climate-explorer-frontend/issues/124
```
rod-glover: Isn't this TODO what you just did here? Or are there more waiting to be extracted?

src/data-services/ce-backend.js
```diff
@@ -59,6 +59,42 @@ function getData(
...
+  //      mean
+  //      median
+  //      stdev
+  //      ncells
+  //      units
+  //      time (median time represented by the dataset)
+  //      modtime (last time dataset was modified)
+  //    }
+  //  }
+  let queryExpString = guessExperimentFormatFromVariable(variable_id, experiment);
```
rod-glover: `const`
Also, wouldn't this be more naturally named `experiment` or `emission`?

src/components/DataControllerMixin/DataControllerMixin.js
corviday: Good catch! One step closer to getting rid of DataControllerMixin entirely. :)

jameshiebert:
Cool, these changes appear to address Rod's concerns. I'm going to go ahead and merge this.

PR #153, opened 2018-05-16
Fix spreadsheet download format

corviday:
File extension for downloading spreadsheets was inconsistently given as both `xlsx` and `xslx` across codebase. Standardized to the correct version, `xlsx`.


jameshiebert:
Oh boy. Good catch! Any evidence of what was the reason for the mass typos?

PR #154, opened 2018-06-08
Bugfixes to handle comparing a climatology with a nominal time dataset

corviday:
Switches to the ce_files dataset, which contains all files from the previously separate `all_CLIMDEX_files `and `all_downscale_files` ensembles. 

Handles the case where a use attempts to compare a climatology dataset (supports annual cycle and long term average graphs) with a nominal time dataset (supports timeseries graphs). Displays the graphs associated with whichever ttime type the primary variable is, but instead of data, puts up an error message about how you can't compare a climatology and a nominal time dataset.

Test docker [here](http://docker-dev02.pcic.uvic.ca:30005).

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -26,17 +26,11 @@ class Header extends React.Component {
                 bsStyle='tabs'
                 justified
               >
-                <LinkContainer to='/climo/all_downscale_files'>
+                <LinkContainer to='/climo/ce_files'>
                   <NavItem eventKey={1}>Standard climatologies</NavItem>
                 </LinkContainer>
-                <LinkContainer to='/climo/all_CLIMDEX_files'>
-                  <NavItem eventKey={2}>ClimDEX climatologies</NavItem>
-                </LinkContainer>
-                <LinkContainer to='/compare/all_downscale_files'>
-                  <NavItem eventKey={3}>Standard comparison</NavItem>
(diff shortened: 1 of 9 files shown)


jameshiebert:
I'm pleased to see how little code it took to make this change. A good indication that we have done our design work properly prior to this point :)

jameshiebert:
From Lee:

> Questions:
> 
>    1. What would we like to name the two tabs? "Explore a dataset" and
>    "Compare datasets" seems clunky, but I haven't thought of anything better.

What about "Single dataset" and "Compare datasets"?

>    2. Would we like to keep the separate extreme precipitation portal?
>    Everything displayed by that portal is now available in the regular
>    comparison portal, but maybe there's some value to a single subject portal
>    making it easier to find a specific set of things?

Yes let's keep it. Slightly more targeted, and it gives us a place to add streamflow when available.

>    3. Would we like to do something to indicate which variables are
>    climatologies and which aren't in the dropdowns, to make it easier for
>    people to find a valid comparison?
> 
>      *  we could colour code variable by which timeset they have
> 
>      *  we could grey out every variable in the secondary variable selection
>    that can't be compared to the primary variable due to its time type

That'd be a good idea. I vote for greying out invalids in the secondary section.

corviday:
A couple more bugs have been found; this branch is not yet ready for merging.

PR #160, opened 2018-06-19
Support selecting a specific variable+description

corviday:
Six of the ETCCDI variables have two different descriptions in the database. Each variable has one description that says `Monthly [Minimum | Maximum] of [thing]` and one that says `Annual [Minimum | Maximum] of [thing]`. The descriptions, at least in the case of these six indices, actually matter: they indicate slightly different ways of calculating an annual version of a monthly index.

The yearly-resolution version of the variable that begins with Annual represents the single highest / lowest value of `[thing]` in a year. The yearly-resolution version of the variable that begins with Monthly represents the mean of twelve values, each of which represents the highest / lowest amount of` [thing]` in one month.

This PR:
- displays each unique variable-description pair in the variable selection dropdown
- keeps track of which description is associated with the variable the user selects
- changes the top level apps (via AppMixin) to filter datasets based on variable-description pairs instead of variable alone

### DIFF
--- a/src/components/AppMixin/AppMixin.js
+++ b/src/components/AppMixin/AppMixin.js
@@ -26,6 +26,7 @@ var AppMixin = {
       ensemble_name: findEnsemble(this.props),
       model_id: '',
       variable_id: '',
+      variable_name: '',
       experiment: '',
       area: undefined,  // geojson object
       meta: [],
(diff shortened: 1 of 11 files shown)


jameshiebert:
Is there a demo available?

corviday:
[Docker demo for this branch](http://docker-dev02.pcic.uvic.ca:30005/)

PR #164, opened 2018-06-25
Flexible map width

corviday:
Changes the width of the mapcontroller div to 100% (of the react layout column), allowing the map to resize if the page size changes instead of covering the graphs. Unfortunately, as far as I can tell, an absolute height must be provided to leaflet, meaning that the map aspect ratio changes based on screen size.

Leaflet's [quickstart guide](https://leafletjs.com/examples/quick-start/) says:
> Make sure the map container has a defined height, for example by setting it in CSS: 
> mapid { height: 180px; }

I wasn't able to get any of the stackexchange workarounds for leaflet's need for an absolute height to work.

[Docker for this branch](http://docker-dev02.pcic.uvic.ca:30005/#/).


jameshiebert:
That looks great! And thanks for pulling the CSS out of the `render()` methods in the components. I was grepping through all of the .css files trying to find where the width was defined and couldn't find it... ugh! This is much better, code-wise, *and* it behaves better too. Thanks!

PR #177, opened 2018-09-14
Set precision for day-counting climdex variables

corviday:
A straightforward change: adds precision values for all climdex variables that measure things as a number of days per year something happens. Trevor has said that fractions of a day are meaningless for these variables, except Tropical Nights (trETCCDI), which should have one decimal place.


rod-glover:
LGTM

PR #178, opened 2018-09-14
Fix test suite after jsdom update

corviday:
See [here](https://github.com/jsdom/jsdom/issues/2304) for discussion.


rod-glover:
Wow, a fix from Dominic Denicola! LGTM.

Issue #175, opened 2018-09-14
Anomalies should be calculated from a historical baseline


jameshiebert:
We talked about this with MoTI, but it seems to have fallen off the RADAR. At present, the "future anomalies" plot calculates the anomaly from a future scenario (2020s) rather than using a historical baseline (1971-2000 or 1981-2010). The latter is the correct way to compute anomalies. This should be fixed before we launch.

corviday:
Is there a preference for which of those two historical climatologies to use?

PR #180, opened 2018-09-17
Improve anomaly graph UI

corviday:
This PR attempts to make the anomaly graph more user friendly by:

1. Changing the display name from "Future Anomaly" to "Change from Baseline"
2. Using percentages for variables for which anomalies are traditionally give in percentages (precipitation) and displaying those percentage changes as a bar graph
3. Using a historical base period, specifically, the most recent climatology that does not include the present year.
4. More succinct axis names

This graph is the most information dense graph we're using, and I've previously reworked a couple times to try and improve readability. I hope this is an improvement.

### DIFF
--- a/src/components/graphs/AnomalyAnnualCycleGraph.js
+++ b/src/components/graphs/AnomalyAnnualCycleGraph.js
@@ -78,7 +78,7 @@ export default function AnomalyAnnualCycleGraph(props) {
     // Select the lowest starting year as the base series for the anomaly graph
     let seriesNames = _.without(graph.data.columns.map(series => _.first(series)), 'x');
     seriesNames.sort();
-    graph = makeAnomalyGraph(seriesNames[0], graph);
+    graph = makeAnomalyGraph(seriesNames[0], props.variable_id, graph);
     return graph;
   }
 
--- a/src/components/graphs/AnomalyAnnualCycleGraph.js
+++ b/src/components/graphs/AnomalyAnnualCycleGraph.js
@@ -3,7 +3,7 @@
  * 
  * Given a data specification (model_id, experiment, variable_id),
  * provides functions to help generate a graph specification contrasting
- * current (2010-2039) and future (2040-2069, 2070-2099) Annual Cycle data:
+ * a baseline historical climatology and future Annual Cycle data:
  * 
  *   - getMetadata locates metadata for multiple three climatology periods
  *   - dataToGraphSpec formats a graph to shade the climatology periods
@@ -29,7 +29,6 @@ export default function AnomalyAnnualCycleGraph(props) {
     //returns metadata for all datasets whose start year is greater 
     //than start, and whose end year is less than end.
     //the interval is exclusive. optionally leave either endpoint undefined.
-    console.log(`${start} ${end}`);
     return _.filter(props.meta, md => {
       return md.model_id === props.model_id &&
       md.experiment === props.experiment &&
@@ -57,7 +56,6 @@ export default function AnomalyAnnualCycleGraph(props) {
     let historicalMetadatas = getDateRangeMetadatas(undefined, currentYear());
     const end_date = _.max(_.pluck(historicalMetadatas, "end_date"));
     historicalMetadatas = _.where(historicalMetadatas, {"end_date": end_date});
-    console.log(`historical end date: ${end_date}`);
     
     //pick the highest-resolution dataset available for that climatology
     const baselineMetadata = _.findWhere(historicalMetadatas, {timescale: "monthly"})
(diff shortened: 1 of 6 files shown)


src/components/graphs/AnomalyAnnualCycleGraph.js
```diff
@@ -25,44 +25,60 @@ import AnnualCycleGraph from './AnnualCycleGraph';
...
+    //returns metadata for all datasets whose start year is greater 
+    //than start, and whose end year is less than end.
+    //the interval is exclusive. optionally leave either endpoint undefined.
     return _.filter(props.meta, md => {
       return md.model_id === props.model_id &&
       md.experiment === props.experiment &&
       md.variable_id === props.variable_id &&
-      md.end_date >= currentYear &&
-      (includeFuture || md.start_date <= currentYear);
+      (_.isUndefined(end) || md.end_date < end) &&
```
rod-glover:
```suggestion
      (_.isUndefined(end) || md.end_date <= end) &&
```

src/components/graphs/AnomalyAnnualCycleGraph.js
```diff
@@ -25,44 +25,60 @@ import AnnualCycleGraph from './AnnualCycleGraph';
...
+      (_.isUndefined(start) || md.start_date > start);
```
rod-glover:
```suggestion
      (_.isUndefined(start) || md.start_date >= start);
```

src/components/graphs/AnomalyAnnualCycleGraph.js
```diff
@@ -25,44 +25,60 @@ import AnnualCycleGraph from './AnnualCycleGraph';
...
 export default function AnomalyAnnualCycleGraph(props) {
   
-  function getPresentMetadatas(includeFuture = false) {
-    // returns an array of all metadata objects that conform 
-    // to the user-selected parameters and contain data that 
-    // includes the current year (or future years, if includeFuture is true).
-    const currentYear = new Date().getFullYear();
-    
+  function getDateRangeMetadatas(start = undefined, end = undefined) {
+    //returns metadata for all datasets whose start year is greater
```
rod-glover:
```suggestion
    // returns metadata for all datasets whose model, experiment, and variable 
    // match the values specified in props, and whose start year is greater 
```

src/components/graphs/AnomalyAnnualCycleGraph.js
```diff
@@ -25,44 +25,60 @@ import AnnualCycleGraph from './AnnualCycleGraph';
...
-  function getPresentMetadatas(includeFuture = false) {
-    // returns an array of all metadata objects that conform 
-    // to the user-selected parameters and contain data that 
-    // includes the current year (or future years, if includeFuture is true).
-    const currentYear = new Date().getFullYear();
-    
+  function getDateRangeMetadatas(start = undefined, end = undefined) {
+    //returns metadata for all datasets whose start year is greater 
+    //than start, and whose end year is less than end.
+    //the interval is exclusive. optionally leave either endpoint undefined.
```
rod-glover: Why is the interval exclusive?

corviday: It was convenient for the [task at hand](https://github.com/pacificclimate/climate-explorer-frontend/issues/175), which was finding a "historical" baseline, ie, one that does _not_ contain the present year.

rod-glover: If it's exclusive on both ends, then a date that exactly a boundary won't get found. Does that matter?

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30501)

src/components/graphs/AnomalyAnnualCycleGraph.js
rod-glover: Nevermind, clearly not. I'm content to leave this as is.

rod-glover:
@corviday , LGTM after updating that comment.

rod-glover:
Also, yes this definitely an improvement to the graph!

src/components/graphs/AnomalyAnnualCycleGraph.js
corviday: At present, it is the desired behaviour both times the function is used:

1. To find the set of all "historical" climatologies, where "historical" is implicitly defined as "all climatologies whose end date is strictly earlier than the present year"
2) having selected a baseline "historical" climatology, to find the set of all "future" climatologies relative to it, where "future" is implicitly defined as "all climatologies whose start year is strictly later than the start year of the base climatology"

I wanted a baseline / future selection process that would reasonably handle the case when the desired baseline climatology (1971-2000) wasn't present. It's not a very complicated algorithm, and if we get overlapping climatologies we may want to revisit these definitions of "historical" and "future."

corviday:
LGTM, thanks for taking over this.

PR #194, opened 2018-09-25
Display datasets with no monthly values in the Long Term Averages graph

corviday:
Handles displaying a set of data with no monthly-resolution data on the Long Term Average graph, such as Heating Degree Days, which is only calculated on an annual basis.

Previously, the Long Term Average Graph default to displaying data from January; if there was no January data, a `long term data unavailable for this model` message was displayed. It now defaults to the first index of the highest time resolution, so either January, Winter, or Annual. If there's not data for any of those three, the error message will still appear.

Also added the props `hideMonths`, `hideSeasons`, and` hideYears` to the TimeOfYearSelector, allowing a component that wants a time selector to get one that contains only resolutions that are actually available in the data.

### DIFF
--- a/src/components/Selector/TimeOfYearSelector.js
+++ b/src/components/Selector/TimeOfYearSelector.js
@@ -49,7 +49,7 @@ class TimeOfYearSelector extends React.Component {
   render() {
     var timesofyear = [];
     function addTimeRange(start, end) {
-      for (const index = start; index <= end; index++) {
+      for (let index = start; index <= end; index++) {
         timesofyear.push([index, timeKeyToTimeOfYear(index)]);
       }
     }
--- a/src/components/Selector/TimeOfYearSelector.js
+++ b/src/components/Selector/TimeOfYearSelector.js
@@ -1,7 +1,7 @@
 /******************************************************************
  * TimeOfYearSelector.js - dropdown for user to select time of year
  *
- * This selector offer a choice of 12 months, 4 seasons, and annual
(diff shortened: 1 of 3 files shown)


rod-glover:
There aren't any degree-day datasets available in the backend yet (I think). Does an annual variable like `prcptotETCCDI` also fall into this category (for testing).

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30502)

corviday:
@rod-glover and I discussed this in person, but for the benefit of my future self looking back at these PRs to write a NEWS.md update: this issue only affects multi-year mean datasets that have only annual resolution.

rod-glover:
@corviday , the essential problem was that the timescales-available and default time of year computations needed to happen in `loadGraph()`, not in `constructor()`. (In `constructor` they only will happen once, at element creation time, and they needed to happen each time the props changed.)

I also removed the monthly-, seasonal-, yearly-data booleans from state. These are values derived directly from props, and unless there is a very good reason, such values should never be put in state, and instead computed where needed. (If the computation is heavy, memoize it.) State is only for things that cannot be computed from existing information, such as user input or the results of fetches.

Finally, I moved the code that computes available timescales and default time of year into util, since they may be used wherever a TImeOfYearSelector element is used. Specifically, they will be part of the fix for the misbehaving ToY selector in the stats summary table.

[Demo](http://docker-dev02.pcic.uvic.ca:30502)

corviday:
LGTM! Thanks for taking over on this.

PR #197, opened 2018-09-25
Fix dataset filter for anomaly graph

corviday:
The filter function for the anomaly graph was using timescale **annual**, which resulted in no data being shown when a dataset had only yearly-resolution climatologies. The correct timescale keyword is **yearly**.


corviday:
... though a yearly-only anomaly graph looks weird and unhelpful. The Long Term Average graph is a much better way to view this information in the degenerate annual-only case.

rod-glover:
Rebased onto master.
Original changes LGTM.

PR #174, opened 2018-09-13
Add guidance and contextual information to CE frontend UI

rod-glover:
# Overview

Addresses #173

1. Replace existing navigation with the following multi-level nav:
    - Home/Data (original content)
       - Single dataset
       - Compare datasets
       - Extreme Precipitation
    - Help
       - FAQ
       - General
    - Science
    - About
      - etc.

2. Add information popovers explaining key elements, including the following:
    - Model, Emission, Variable selectors
    - Graph tabs

Revamping the navigation includes the following changes:

- Upgrade `react-router` to v4, "dynamic routing." This is a significantly different, more flexible, and (amazingly) at least as easy to use package. Worth reading about, and it works better than our old routing did.
- Use React Bootstrap `Nav` components throughout.
- Place top-level nav in app header; secondary nav appears in the page.
- Add a marmot. (Image by permission from [V.I. Marmot Recovery Foundation](https://marmots.org/)!)

Additionally, I have revised the layout to make it more consistent from data view to data view, to label things more clearly (with info popups for further explanation), and to indicate hierarchy and scope.

# To do

## Functionality

- [x] Add Datasets Summary panel to improve user's understanding of what dataset filtering does

## Layout

- [x] Fix layout of selectors and buttons in graphs
- [x] Fix layout of selectors and buttons in stats table
- [x] Fix layout of stats table proper
- [x] Add flow arrows to show relationship of datasets to data presentation elements

### FAQ

- [x] FAQ feedback request with contact info

### General

  - [x] What data is available
  - [x] Meanings of variables
  - [x] Map tools
  - [x] Map scales

### ~~Glossary~~

  - [ ] ~~Model, Emission Scenario, Variable~~
  - [ ] ~~Codes: do they belong in Glossary??~~
    - [ ] ~~Model codes~~
    - [ ] ~~RCP codes~~
    - [ ] ~~Variables, i.e., glossary item per variable, e.g., precip, climdex variables, ...~~
 - [ ] ~~Graph types (e.g., Long Term Average) - can probably dup from info popups~~

### Climate Explorer

- [x] Explain better
- [x] Author
- [x] Copyright?
- [x] Licence?
- [x] Terms (see PCIC Data Portals)

### Credits

- [x] Links to institutions providing data

### Contact

- [x] Feedback on tool
- [x] Scientific questions
- [x] PCIC

## Info popups

- [x] Map Settings
- [x] Map tools: Link in title popup to Help
- [x] Map scales: Link in title popup to Help
[...]

Issue #173, opened 2018-09-04
Add guidance and contextual information to CE frontend UI


corviday:
Possible inclusions:

- [ ] general help and usage
- [ ] scientific references
- [ ] credits and acknowledgements, especially MOTI
- [ ] per-element help
- [ ] FAQ
- [ ] About page
   - [ ] version
   - [ ] release
   - [ ] how to get help
   - [ ] link to github issues page

PR #199, opened 2018-10-03
Guidance, Part 1: Revise navigation

rod-glover:
Issue #173 Guidance
Part 1: Revise Navigation (cherry picked from humongous PR #174)

Replace existing navigation with the following multi-level nav:

- Home/Data (original content)
   - Single dataset
   - Compare datasets
   - Extreme Precipitation
- Help
   - FAQ
  - General
- Science
   - About
      - etc.

Revamping the navigation includes the following changes:

- Upgrade `react-router` to v4, "dynamic routing." This is a significantly different, more flexible, and (amazingly) at least as easy to use package. Worth reading about, and it works better than our old routing did.
- Use React Bootstrap `Nav` components throughout.
- Place top-level nav in app header; secondary nav appears in the page.
- Add a marmot. (Image by permission from [V.I. Marmot Recovery Foundation](https://marmots.org/)!)


rod-glover:
I'll put up a docker for this tomorrow first thing.

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30501) of this branch up on docker

It contains only navigation components, but those should look exactly the same as the [demo](http://docker-dev02.pcic.uvic.ca:30500) for the all-in-1 branch.

corviday:
The demo looks very good, just starting to look over the code now.

src/components/App/__x_tests__/smoke.js
```diff
@@ -0,0 +1,17 @@
+// Cannot get this test to pass. Hence the renamed __tests__ directory.
+// There is a problem with C3; see https://github.com/c3js/c3/issues/2129
+// but the recommended solutions do not work. Preserving this attempt in the
+// hope that some future effort can actually make it work.
+
```
corviday: There are several tests I've previously been unable to get to work, [like this one](https://github.com/pacificclimate/climate-explorer-frontend/blob/22c63d4f2ad220ab83dd5336e67e12bf835b8a22/src/core/__tests__/util-test.js#L192). I like the **xit** function for this because sometimes tests are a nice bit of documentation, so it can be useful to leave them findable in the normal location.

src/components/navigation/NavRoutes/NavRoutes.js
```diff
@@ -0,0 +1,133 @@
...
+        {item.label}
+      </NavItem>
+    </LinkContainer>
+  );
+
+  const routes = navSpec.items.map((item, index) =>
+    <Route
+      key={index}
+      path={withBasePath(item.subpath)}
+      component={item.component}
```
rod-glover: Verify that this is correct; should it be `navSubpath || subpath`?

src/components/App/__x_tests__/smoke.js
rod-glover: I wanted to use `xit` too, but the error occurs before the test -- during the phase when the component under test is being imported. Hence this.

corviday: That makes sense.

PR #200, opened 2018-10-05
Guidance, Part 2: Information popovers

rod-glover:
Issue #173 Guidance
Part 2: Information popovers (cherry picked from humongous PR #174)

Add information popovers to various components of the application that explain context, meaning, usage, etc. 

Summary:

- Popovers are normally hidden, and are triggered by clicking on the info (circle-i) icon. An open popover is closed by clicking anywhere outside it.

   It would be better if the popover were triggered by hovering over the icon, but that cannot be combined with click outside to close, and the latter is more important. We can almost certainly fix this with a custom overlay component, but that seemed too time consuming for this round of changes.

- Because the info popovers provide much more context, some of the labelling in Map Settings (and a couple of other places) has been simplified. 

- Two new components, `Information` and `LabelWithInfo` have been defined to encapsulate the info popovers and their inclusion into labels (which is by far the most common usage for them).

- Labels for all kinds of things from panel headings to graph tabs to dropdowns have been pulled out into `InformationItems.js`, where all such items are all centrally defined and can share common elements. This file exports a bunch `Information` and `LabelWithInfo` instances.

- A couple of components have been renamed with more informative, explicit names.


src/components/MapSettings/TimeSelector.js
```diff
@@ -53,16 +50,18 @@ export default class TimeSelector extends React.Component {
       return <NullTimeSelector/>;
     }
 
+    const label =
+      <TimeSelectorLabel temporalLabelPart={this.temporalLabelPart()}/>
+    ;
```
rod-glover: Possibly inline  `this.temporalLabelPart()`. It doesn't really serve any purpose being pulled out as a function: it's evaluated once, here.

src/components/app-controllers/SingleAppController/SingleAppController.js
```diff
@@ -36,52 +40,52 @@ export default createReactClass({
...
-    return !(datafile.multi_year_mean == false && datafile.timescale == "monthly");
+    return !(datafile.multi_year_mean === false && datafile.timescale === 'monthly');
   },
 
   //Returns metadata for datasets with thethe selected variable + scenario, any model.
   //Passed as a prop for SingleDataController to generate model comparison graphs.
   getModelContextMetadata: function () {
     return _.where(this.state.meta,
-        {variable_id: this.state.variable_id, experiment: this.state.experiment});
+        { variable_id: this.state.variable_id, experiment: this.state.experiment });
```
rod-glover: Sorry, several changes in this file are due to automatically complying with our standard ESLint conventions, and really have nothing to do with this PR's actual purpose. They can be removed if you find them too distracting.

src/components/app-controllers/SingleAppController/SingleAppController.js
```diff
@@ -36,52 +40,52 @@ export default createReactClass({
...
+          <Col lg={4} md={4}>
+            <VariableDescriptionSelector
+              label={variableSelectorLabel}
+              onChange={this.handleSetVariable.bind(this, 'variable')}
+              meta={this.state.meta}
+              constraints={{ model_id: this.state.model_id }}
+              value={_.pick(this.state, 'variable_id', 'variable_name')}
+            />
+          </Col>
         </Row>
```
corviday: If you change the order of the selectors (swapping Variable and Emissions Scenario) you may want to change the "hierarchy" of the selectors as well to make it more intuitive.

Previously, the selectors were arranged model -> variable -> emissions from left to right. Model allowed the user to select any models. Variable allowed the user to select only variables available for the selected model. Emissions allows the user to select only emissions available for the selected model and variable. There was an implicit left to right hierarchy to the selectors such that each selector would disable all items that didn't make sense based on the "already made" choices in selectors to the left.

Now going from left to right we have model -> emissions -> variable, but we've left the item-availability hierarchy model -> variable -> emissions.

corviday: ... though switching the hierarchy to model -> emissions -> variable gives you a different UI awkwardness: the user has to guess the "right" emissions scenario before they can select the variable they want, and if they select the "wrong" emissions scenario, the variable they want will be disabled. 

That setup was unfortunately necessary for the dual-variable portal, since the two variable are required to share a model and emissions scenario. I assume you switched the order to make the portals more consistent, which makes sense.

What the user really cares about is probably mostly the variable. If we had only this portal, it might make most sense to set up the selectors and hierarchy variable -> emissions scenario -> model, but there's some value in consistency across portals.

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+
+
+///////////////////////////////
+// Dataset filters (Model, Emission, Variable)
+///////////////////////////////
+
+export const datasetFilterPanelLabel = (
+  <LabelWithInfo label='Dataset Filter'>
+    <p>
+      This is the primary filter for the datasets to be displayed.
```
corviday: Maybe switch "This is the primary filter for the datasets to be displayed" to "This is the primary filter used to select datasets to display"? The "for" seems unclear as to what the relationship between filter and datasets is.

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+  </LabelWithInfo>
+);
+
+export const modelSelectorLabel = (
+  <LabelWithInfo label='Model'>
+    <p>
+      GCM model with which the climate data was generated.
+    </p>
+    <p>
+      Models are identified by short codes. For full model identification,
```
corviday: Maybe add a single sentence about what a model _is_, like "A model is a thermodynamics simulation of climate processes in the atmosphere and on earth's surface" or something?

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    <p>
+      On the graph(s), the Dataset selector appears directly above the graph.
+    </p>
+    <p>
+      Datasets are identified by a combination of
+      model run id (e.g., <code>r1i1p1</code>) and
+      averaging period (e.g., <code>1961-1990</code>).
+    </p>
+  </LabelWithInfo>
+);
```
corviday: Ugh, that's a lot of text. But I agree it's all necessary.

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+const LeafletControlContainer = ({ children }) => (
+  <span className='leaflet-touch'>
+      <span className='leaflet-control-container'>
+        {children}
+      </span>
+    </span>
+);
+
+const mapSettingsControl = (
+  <span>
```
corviday: This doesn't seem to be used anywhere; I'm guessing it'll make more sense in a later PR and is just an artifact of having to split up the big PR.  I'm not entirely sure about the reasoning for including a control in the file-of-help-text, but I imagine it'll be clearer later.

corviday: Or, wait, is it just meant to be a representation of the Map Settings button, not the actual button, and its only use is in other help elements? In that case, could it get a more descriptive name, like mapSettingsControlRepresentation, or ... uh ... something shorter than that.

corviday: Okay, I see that it's used in help elements in the massive uni-branch, and was just a victim of the Great Untangling. Never mind!

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+
+const downloadFormats = (<span>
+  Data may be downloaded in Microsoft Excel compatible format (XSLX)
+  or in Comma Separated Values (CSV) format.
+  For details on these formats, see <Link to='/help/general'>Help</Link>.
+</span>);
+
+const exportDataLabel = 'Export Data';
+
+export const downloadGraphDataLabel = (
```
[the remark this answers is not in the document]
corviday: I see this is in the massive Uni-branch. Never mind.

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+      <li>Click on the sort column to change the direction of sorting.</li>
+    </ul>
+
+
+
+    .
+  </LabelWithInfo>
+);
+
+export const exportStatsTableDataLabel = (
```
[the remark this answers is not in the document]
corviday: I see it is used in the massive uni-branch and its purposeless here is just a side effect of the Great Untangling. Never mind!

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    <p>{annualCycleGraphDefn}</p>
+    <p>{spatialAveragingDefn}</p>
+    <p>{datasetSelectorDefn}</p>
+  </LabelWithInfo>
+);
+
+const ltaGraphDefn = `
+  A long term average graph presents average values of the selected variable
+  for the selected month, season, or the year, averaged
+  over several different multi-decade period.
```
corviday: "over several different multi-decade period" -> "over several different multi-decade periods"

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    <p>{spatialAveragingDefn}</p>
+    <p>{datasetSelectorDefn}</p>
+  </LabelWithInfo>
+);
+
+const ltaGraphDefn = `
+  A long term average graph presents average values of the selected variable
+  for the selected month, season, or the year, averaged
+  over several different multi-decade period. 
+  There is one data point per mulit-decade averaging period.
```
corviday: "mulit-decade"->"multi-decade"

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    </p>
+  </LabelWithInfo>
+);
+
+///////////////////////////////
+// Graphs
+///////////////////////////////
+
+const annualCycleGraphDefn = `
+  An annual cycle graph presents spatially averaged values of a multi-year
```
corviday: I wonder if the first "An" should be a "The" or "This"  - is what we're saying here true of all annual cycle graphs in general, or just the ones in CE?

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    Annual cycle graphs showing the yearly, seasonal, and monthly
+    mean values of the two selected variables (if different).
+    <p>{annualCycleGraphDefn}</p>
+    <p>{spatialAveragingDefn}</p>
+    <p>{datasetSelectorDefn}</p>
+  </LabelWithInfo>
+);
+
+const ltaGraphDefn = `
+  A long term average graph presents average values of the selected variable
```
corviday: Similarly, I wonder if the "A" here should be a "The" or "This".

src/components/guidance-content/info/InformationItems.js
```diff
@@ -0,0 +1,617 @@
...
+    <p>
+      This graph shows the influnce of the secondary variable
+      on the primary variable irrespective of time.
+    </p>
+    <p>
+      It is composed from timeseries data with matching availability for
+      both variables. Each point in time {t} with data from each variable
+      ({t}, {V1}({t})) and ({t}, {V2}({t}))
+      appears as the scatterplot point
+      ({V2}({t}), {V1}({t})).
```
corviday: Nice job on the formatting here!

corviday:
It looks great! Some minor questions or quibbles, but nothing major.

src/components/guidance-content/info/InformationItems.js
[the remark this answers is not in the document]
rod-glover: Yes, sorry about that. I grew weary of horrible merges for this file and started just replacing it with the final result everywhere.

src/components/guidance-content/info/InformationItems.js
rod-glover: Even if it's just the ones in CE, we present several. I don't know whether there is a general notion of annual cycle graph in common use, but I intended this to describe our coinage.

rod-glover:
@corviday , LGTY?

corviday:
Experiment selection needs to be _not_ filtered on state.variable_id (line 57). Other than that, LGTM!

Confusingly, emissions scenario and variable use two completely different available-item filtering mechanisms. Experiment and model both just use _.filter(), but variable is not actually a single attribute, and has a dedicated selector class to handle its formatting and filtering complexities.

It occurs to me now, belatedly, that it would improve readibility to create `ModelSelector` and `EmissionsSelector` classes to match the existing VariableSelector class, and have all of them accept the same kinds of props: a list of metadata objects and a constraint object, so that they all use the same filtering mechanism as far as the components containing them are concerned.

rod-glover:
Aha, I was wondering where that was hiding. Thanks. 

I agree re. your suggested change and put it into issue #202.

PR #210, opened 2018-10-16
Guidance, Part 6: Miscellaneous

### LINKED ISSUE #204: Navigation URLs are wrong when app root URL contains path
I thought we fixed this a while back, but I can find no trace of either the issue/discussion or the code change for it (which I thought was a tricky setting in `package.json`). 

The problem as we see it now is likely due to different behaviour of v4 of `react-router`, which we recently adopted.

rod-glover:
Issue #173 Guidance
Part 6: Miscellaneous cleanup and formatting, navigation fix PR #174)
This PR is the last of the Guidance series (for now).
Resolves https://github.com/pacificclimate/climate-explorer-frontend/issues/173
Resolves #204 

This PR includes the following:

- Format nav bars a little differently.
- Clean up code; specifically the graph tabs code.
- Clean up some wording in help/info.
- Fix navigation: enable us to deploy to an arbitrary URL whose base path is defined in the env. variable `CE_BASE_PATH`. Yay!

Please forgive the horrible commit history. Rushing a bit.


rod-glover:
@corviday , ready when you are. Tomorrow AM is fine for me.

corviday:
A mild bit of navigational weirdness, may not be important enough to fix:

In most cases, if you are looking at a specific portal, then click on help, then click on home/data to return to the portal, you will find yourself at the same portal you started at.

However, if you follow the fourth link to the compare portal, then view help, then return to the portal, you will be at the single portal.

If, on the other hand, you follow one of the first three links to the single portal, then switch to the compare portal, opening and closing help will return to the compare portal, as expected.

corviday:
That weird blip is the only navigation difficulty; otherwise every seems to work great.

src/components/App/App.js
```diff
@@ -67,13 +67,62 @@ export default class App extends React.Component {
...
+    // 1. Configure the server to ignore the client-side routing
+    // part of the URL (i.e., to care only about the base path). See the
+    // create-react-app documentation
+    //    (https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#serving-apps-with-client-side-routing).
+    //
+    // 2. Use a hash (#) to separate the base path from the client-side routes.
+    //    This makes the client-side route component not a part of the URL path.
+    //
+    // We choose option 2, hence the hash (#) following the base path in the
+    // `Router.basename` value below.
```
corviday: I appreciate this detailed comment!

Has the default deployment procedure of a development docker changed at all, such as needing to manually supply a new environment variable? If so, could you edit the main readme to reflect that?

src/components/data-controllers/SingleDataController/SingleDataController.js
```diff
@@ -177,6 +177,37 @@ export default createReactClass({
       this.state.dataTableTimeOfYear
     );
 
+    // Spec for generating tabs containing various graphs.
+    // Property names indicate whether the dataset is a multi-year mean or not.
+    const graphTabsSpec = {
```
corviday: This is pretty neat! Is the plan to do it in all the datacontrollers eventually, given time enough?

rod-glover: Heh, no time like the present.

rod-glover:
I've added issue #213 for the nav difficulty. It's an initialization problem. Fix later.

rod-glover:
@corviday , thanks for your very careful exploration of navigation. Good catch!

PR #228, opened 2018-11-20
Set precision for *all* variables, resolve #223

### LINKED ISSUE #223: Too many significant figures in value displays
Feedback from Trevor Murdock.

Many of the value displays show 4 significant figures (for precipitation variables, at the least). This implies a precision far beyond what is scientifically defensible. No more than 2 sig figs should be shown for any model-derived data value.

corviday:
Sets precision for all current variables.

### DIFF
--- a/variable-options.yaml
+++ b/variable-options.yaml
@@ -65,6 +65,9 @@ csdiETCCDI:
 cwdETCCDI:
   decimalPrecision: 0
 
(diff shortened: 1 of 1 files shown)


rod-glover:
I checked against the list in the Variables selector, and all are now specified.
LGTM.

PR #234, opened 2018-11-22
Show dataset selection in Data Map panel label

### LINKED ISSUE #222: Data Map panel label should show dataset selection within map
In other words, the panel label should be like the legend in the map, which indicates the dataset (time of year) selection made in Map Settings. If that can't be accomplished easily, it should be removed from the header altogether. If it can be, remove the legend from the map.

rod-glover:
Resolves #222 .

This PR makes the map legend (displayed on the map proper) and the Data Map panel label the same.

I'm leaving the legend on the map because if we restore map printing, then it will be useful to have it. It can be removed easily if it is genuinely unwanted.

The legend is somewhat modified from the original. It may be a little too verbose, but it contains complete information about the dataset selection, which the original label did not. It also contains an annotation as to which variable is the raster and which is the isolines. This may be an annotation too far.

The map legend and the (similar) model-emissions-variable summary titles now both display "Loading..." when there is no valid data to fill them out with. This is much better looking than a string full of "undefined" for this case, and it improves the user experience because it indicates what is actually happening.

### DIFF
--- a/src/components/MapFooter/package.json
+++ b/src/components/MapFooter/package.json
@@ -1,6 +1,0 @@
-{
-  "name": "MapFooter",
-  "version": "0.0.0",
-  "private": true,
-  "main": "./MapFooter.js"
-}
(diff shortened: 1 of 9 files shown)


rod-glover:
Single variable case
image

Dual variable same case
image

Dual variable different case
image

corviday:
Could you put up a demo?

rod-glover:
Shortened message

PR #236, opened 2018-11-29
Fix imported polygon behaviours

### LINKED ISSUE #185: Cannot edit or delete an imported polygon
To reproduce:

1. Import a GeoJSON polygon.
2. Click Edit Polygon button. Nothing happens.
3. Click Delete Polygon button. Nothing happens.

### LINKED ISSUE #231: Delete polygon behaviour is highly counterintuitive and appears buggy
1. Cannot delete a single polygon by clicking on it. To reproduce:
   1. Draw a polygon
   1. Select Delete Polygon tool
   1. Move mouse over polygon
   1. Mouse pointer changes to select pointer (finger)
   1. Click on polygon to try to delete it
   1. Polygon is not deleted; no feedback of any kind is given
   1. Bonus weirdness follows:
   1. Click Delete Polgyon > Save
   1. Polygon remains (boo)
   1. Select Delete Polygon tool (again)
   1. Point at polygon and click
   1. Polygon is deleted, this time

1. Cannot delete the first of multiple drawn polygons. To reproduce:
   1. Draw two or more polygons
   1. Select Delete Polygon tool
   1. Point at first polygon drawn and click
   1. Nothing happens
   1. Point at any of the other polygons, in any order, and click
   1. Polygon is deleted
   1. After deleting all but first polygon, still can't delete first one
   1. Bonus weirdness follows:
   1. Click Delete Polgyon > Save
   1. Only first polygon remains (i.e., deleted polygons stay deleted; yay!, sort of)
   1. Select Delete Polygon tool (again)
   1. Point at first polygon drawn and click
   1. Polygon is deleted, this time

Notes:

1. All polygons can be deleted in all cases by selecting Clear All from Delete Polygon popout menu.

rod-glover:
Resolves #185 
Resolves #231 

There were two intertwined problems in these two issues, so they are solved jointly in this PR:

1. A combination of older versions of Leaflet, Leaflet Draw, React, React Leaflet, and React Leaflet Draw yielded buggy behaviour (the mysterious persistent polygon).

1. The import functionality did not inject the imported geometry into the Feature Group that the Edit control affects. Therefore once a polygon was imported, it could never be deleted or otherwise modified.

For details on these problems, see the issues mentioned. The solutions are as follows:

1. Update Leaflet, Leaflet Draw, React, React Leaflet, and React Leaflet Draw to latest stable versions. There are few code consequences of this, save the need to inject the `leaflet` prop into our custom React Leaflet components using the new HOC `withLeaflet` now supplied with React Leaflet.

1. This is one of a couple of solutions considered in #185, and significantly the simpler (if perhaps less elegant) of the two: 
    1. Create an extended `FeatureGroup` component (`LayerControlledFeatureGroup`) that is controlled with respect to its layer content. Control that content in `DataMap` (i.e., by a `DataMap` state item). 
    1. Update that state from both the edit tool and the geometry import tool. 
    1. Move the geometry import/export controls from the map controllers into `DataMap`. They belong more naturally there, reduce redundant code, and save us from having to manage the geometry later state higher up the hierarchy. Win all round.

Notes to reviewers: 

1. The only significant changes to the code are in the DataMap component. 
1. `LayerControlledFeatureGroup` is new but fortunately very simple.
1. The rest is housekeeping and a lot of comments.


rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30501)

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30501) updated

corviday:
I don't think I'd seen the little popup that tells you the area of the rectangle you're drawing before, that's _very_ cool! 

Can the units be easily changed to square kilometers? I don't think this is a priority, so skip it if it's a lot of work, but if it's just setting an attribute or something, maybe consider km instead of ha, since other displayed units use km?

src/components/DataMap/DataMap.js
```diff
@@ -181,26 +285,36 @@ class DataMap extends React.Component {
...
+            title='Import polygon'
+          />
+        </StaticControl>
 
-        {
-          this.props.area &&
-          <GeoJSON
-            data={this.props.area}
+        <StaticControl position='topleft'>
+          {/* See comments above regarding current GeoExporter arrangement. */}
```
corviday: Which comments is this referring to? I was hoping for an explanation of the empty {}, but did not find one.

src/components/NcWMSColorbarControl/NcWMSColorbarControl.js
```diff
@@ -1,13 +1,14 @@
...
+import { MapControl, withLeaflet } from 'react-leaflet';
 
 import LeafletNcWMSColorbarControl from './LeafletNcWMSColorbarControl';
 
 
 // TODO: https://github.com/pacificclimate/climate-explorer-frontend/issues/123
 class NcWMSColorbarControl extends MapControl {
   createLeafletElement(props) {
+    console.log('NcWMSColorbarControl.createLeafletElement', props)
     return new LeafletNcWMSColorbarControl(props);
```
corviday: Perhaps remove debug output?

src/core/geoJSON-leaflet.js
```diff
@@ -0,0 +1,90 @@
...
+  // We handle (a) by returning an array of layers, one for each Feature in a
+  // FeatureCollection, and an array of one item for a single Feature.
+  //
+  // We handle (b) similarly to (a).
+  //
+  // We handle (c) by ignoring it for the moment, since it is unlikely to arise
+  // and in any case ought to be handled OK by Leaflet (since `FeatureGroup` is
+  // a subclass of `Layer`).
+
+  console.log('geoJSONToLeafletLayers geoJSON =', geoJSON);
```
corviday: Possibly remove debug output?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "prestart": "if [ ! -f ./variable-options.yaml ]; then touch ./variable-options.yaml; fi",
-    "start": "CE_CURRENT_VERSION=$(./generate-commitish.sh) webpack-dev-server --host 0.0.0.0 --history-api-fallback",
+    "start": "CE_BACKEND_URL=https://services.pacificclimate.org/marmot/api NCWMS_URL=https://services.pacificclimate.org/marmot/ncwms TILECACHE_URL=https://tiles.pacificclimate.org/tilecache/tilecache.py CE_ENSEMBLE_NAME=all_downscale_files CE_CURRENT_VERSION=$(./generate-commitish.sh) webpack-dev-server --host 0.0.0.0 --history-api-fallback",
```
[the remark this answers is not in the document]
rod-glover: So that I have some data to work with. The previous case defaults to non-existent data sources. The new one uses the production data sources, which is what we ought to be using unless there is some special case we are testing with different data sources.

src/components/DataMap/DataMap.js
```diff
@@ -58,16 +127,19 @@ class DataMap extends React.Component {
...
-        // Passing a bounding box with identical eastmost and westmost bounds to
-        // ncWMS results in an error, so create a new Canada-only bounding box and
-        // ignore the worldwide extent of this map.
+        // are both zero (actually -.5675 or something like that, the center
+        // of a cell with one edge at 0.)
+        // Passing a bounding box with identical eastmost and westmost bounds
+        // to ncWMS results in an error, so create a new Canada-only bounding
+        // box and ignore the worldwide extent
+        // [[-122.949219,63.632813],[-113.769531,68.222656],
+        // [-110.742187,63.242187],[-122.949219,63.632813]]
```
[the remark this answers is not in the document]
rod-glover: These were there originally. This comment is confused/ing but I couldn't figure out how to fix it without a lot of time-consuming research, so I just left it as it is. These comments have been reformatted (line length) to pass ESLint.

src/components/DataMap/DataMap.js
rod-glover: The only way to insert a comment inside a JSX component declaration is to start a code section (`{}`) that contains the comment(s). It's awkward, but it's better than no comments at all. So this comment refers to the next line

rod-glover:
Unfortunately, the area popup is only available for the rectangle, and the perimeter popup only available for the polygon, and neither show when editing one or the other, so I disabled them all. Sigh.

All other comments addressed by replies or modifications.

PR #240, opened 2018-12-04
Limit geometry to one active polygon

### LINKED ISSUE #184: Do we handle multiple polygons?
The map tooling allows drawing any number of polygons. Do we use all of them?

If the answer is no: Should we allow multiple polygons and use all of them?

If we should allow and use only one polygon, then fix this feature so that only one can be drawn.

rod-glover:
Resolves #184 

This PR implements the following behaviours:

- Allow only one geometry (polygon) to be drawn. 
- Disable drawing controls and import control when one or more geometries are present.
- If more than one geometry  is imported (e.g., from a file containing a GeoJSON FeatureCollection), then only the first (in order of importation) will be "active", meaning that only the first geometry will be used to form spatial data averages. 
- Color inactive geometries grey (as opposed to default blue for active geometry). This is simple to implement and avoids confusing users with the effect of importing only part of a file.
- Polygon export tool exports only the active geometry.
- Disable polygon export tool when no geometries are present.

### DIFF
--- a/src/components/DataMap/DataMap.js
+++ b/src/components/DataMap/DataMap.js
@@ -301,22 +301,15 @@ class DataMap extends React.Component {
           {...this.props.isoline}  // update when any isoline prop changes
         />
 
-        <StaticControl position='topleft'>
-          {
-            allowGeometryDraw &&
-            <GeoLoader
-              onLoadArea={this.handleUploadArea}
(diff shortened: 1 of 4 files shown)


rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30502)

PR #245, opened 2018-12-06
Fix time of year selection in LTA graph

### LINKED ISSUE #244: Cannot select time of year in LTA graph
The selector does not change when an item is selected.
This is directly related to the fact that the default time for the dataset is *always* set when the view is updated. Who coulda done that???

rod-glover:
Resolves #244 

Also makes component lifecycle (a) up to current React lifecycle API and recommendations, and (b) considerably simpler. Async data wrangling is correspondingly simplified too.

Note to reviewer: I'm afraid that the reorganization of component `LongTermAveragesGraph` has made the diff a bear. Given that it really is a reimplementation, possibly just review the new code without the diff.

### DIFF
--- a/src/components/graphs/LongTermAveragesGraph/LongTermAveragesGraph.js
+++ b/src/components/graphs/LongTermAveragesGraph/LongTermAveragesGraph.js
@@ -1,3 +1,15 @@
+// This component renders a complete long-term average graph, including a
+// selector for the averaged period of interest (e.g., January, Fall, Year).
+// A long-term average graph presents spatially averaged values of
+// climatological average data for the selected model, variable,
+// experiment, and the locally-selected period of interest,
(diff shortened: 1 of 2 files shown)


corviday:
The demo seems to crash spectacularly if I switch to a non-MYM dataset, like growing season ength, or cold spell duration.

rod-glover:
Spectacularly! That's what I strive for! :stuck_out_tongue_winking_eye: 

It was due to adopting the (otherwise spot-on) React data-fetching pattern without quite enough thought. Easy fix.

corviday:
Excellent! I'll give it another look.

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30504) updated

Issue #250, opened 2018-12-10
Export to shapefile creates invalid shapefile content


rod-glover:
To reproduce:

1. Draw a polygon.
1. Export to shapefile.
1. Import to [Mapshaper](https://mapshaper.org/) (online editor for map data).
1. Error message: "Error importing POLYGON.shp Counting error"
1. Unsurprisingly, our GeoImport component cannot import these shapefiles either. But it can import valid shapefiles.

PR #251, opened 2018-12-10
Save zipped shapefiles to root folder in zipfile

### LINKED ISSUE #239: Exported zipped shapefile should have files at root
Currently, the exporter exports to a zip file containing a directory `layers` containing the expected .shp, .shx, .prj, and .dbf files. It should not do this, and instead should place these files at the root of the zipfile.

rod-glover:
Resolves #239 

Creates an alternate version of the `download` function of the `shp-write` package. Pretty simple code.

### DIFF
--- a/src/core/geo.js
+++ b/src/core/geo.js
@@ -5,7 +5,67 @@ import { parse, stringify } from 'wellknown';
 import _tokml from 'tokml';
 import _togpx from 'togpx';
 import shp from 'shpjs';
-import { download } from 'shp-write';
+import { write } from 'shp-write';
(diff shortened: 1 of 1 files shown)


rod-glover:
Sure, but don't spend a ton of time on it. The import/export tool may change quite a bit soon as we fix the import (mainly) bugs. It's pretty sketchy right now, though I believe export is reasonably solid. Thanks for offering.

rod-glover:
@corviday : See issue #250 . There are known bugs here. If you wish to explore further and add to that issue, go ahead.

Issue #219, opened 2018-11-15
Switching variables may result in invalid selected climatologies


corviday:
To reproduce:

1. Select a dataset that has a 1961-1990 climatology, such as CanESM, rcp85, tasmax.
2. Switch to a dataset that _doesn't_ have a 1961-1990 climatology, such as CanESM, rcp85, cdd.

The 1961-1990 climatology will remain selected, despite not being present in the new data selection:
* climatology selector for relevant graphs will display `{"start_date":1961, "end_date": 1990, "run": "r1i1p1"}` 
* graphs will display "No data to show" error message

This is a newly-uncovered problem that affects the Degree Day and Return Period data, which are available for a different set of climatologies than the Climdex and Downscaled data. These datasets have only 4 climatologies each.

rod-glover:
Does our current code (attempt to) preserve the climatology selection between datasets? It would seem so.

If so, the questions are 

1. Is this preservation desirable?
1. Would it be better to reset the climatology selection (to, say, the earliest one available for the dataset) whenever a different dataset is selected?
1. If there is a convincing argument for preserving selection, then we need to do the item above when there is no matching climatology for the selected dataset.

rod-glover:
Closed duplicate

rod-glover:
Closed wrong one

corviday:
The current code preserves the selected climatology pretty much by default: it just doesn't update the chosen period in state when the user switches variables. I think that's reasonably intuitive behaviour, though it needs patched to handle the case when there isn't a matching climatology. 

I could also see an argument for always displaying the earliest period when switching variables, or always displaying the current period when switching variables.

I don't think users would have trouble following any of those strategies, whichever we decided to apply.

PR #255, opened 2018-12-11
Fix invalid default climatology (Dataset) selection

rod-glover:
Resolves https://github.com/pacificclimate/climate-explorer-frontend/issues/219

This PR follows follows the recommended lifecycle hook usage for asynchronous data fetching, and adapts it to the case where data fetches are triggered by state change as well as by props change. This proved somewhat tricky, but it's logical once understood.

Note to reviewer: Reorganization of component AnnualCycleGraph has made the diff a bear. Given that it really is a reimplementation, possibly just review the new code without the diff.

### DIFF
--- a/src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
+++ b/src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
@@ -10,15 +10,15 @@ import ExportButtons from '../ExportButtons';
 import { exportDataToWorksheet } from '../../../core/export';
 import { getTimeseries } from '../../../data-services/ce-backend';
 import {
+  defaultDataSpec,
   validateAnnualCycleData,
   validateUnstructuredTimeseriesData,
 } from '../../../core/util';
 import {
-  displayError,
   noDataMessageGraphSpec,
-  blankGraphSpec,
   multiYearMeanSelected,
-  shouldLoadData,
+  errorMessage,
+  loadingDataGraphSpec,
 } from '../graph-helpers';
 
 // This component renders an annual cycle graph, including a selector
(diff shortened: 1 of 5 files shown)


rod-glover:
Oops, broken. Don't review yet.

rod-glover:
demo shortly

rod-glover:
hmmm, dockerhub isn't triggering the build

rod-glover:
manually triggered a build, wonder what is wrong

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30506/#/data/climo/ce_files)

src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
```diff
@@ -129,28 +205,38 @@ export default class AnnualCycleGraph extends React.Component {
     exportDataToWorksheet(
       'timeseries',
       _.pick(this.props, 'model_id', 'variable_id', 'experiment', 'meta'),
-      this.state.graphSpec,
+      this.graphSpec(),
```
corviday: So the graph is rebuilt from API results if someone wants to download a file? Have you tested about how long that takes? I assume it's pretty fast - the slow part is fetching the data in the first place.

corviday:
Code looks good, is tehre an updated demo, or should I use the previous one?

rod-glover:
You rock, Mr. Zeman.
Updated demo coming in about 3 minutes.

rod-glover:
Argh, I also have to rebase or merge master into this baby. But I'll do that later.

corviday:
It looks like the conflict is a single blank line in chart-generators, shouldn't be too bad. :)

rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30506/#/data/climo/ce_files) updated

rod-glover:
Well that's just .... awesome.

corviday:
Demo looks good, I can't seem to break it.

I wonder whether "always display the earliest available climatology when the user switches datasets" is the best strategy, maybe "always display 'today'" or "try to match the previously selected climatology if possible" would be more intuitive, but I think we can wait to see if users suggest something - this solves the presenting bug, and it's certainly not difficult to understand.

src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
rod-glover: Good question. Happily, by design, no backend API calls are made:
`graphSpec` (when it has data) calls 
- `this.getMetadatas`, which in turn calls `props.getMetadata`, which is a function that returns a filtered subset of the already-fetched metadata. 
  - The components which specialize ACG (by passing in `props.getMetadata`) do not (or at least should not) hit the backend to obtain this metadata; they already have it (passed in as a prop). 
  - Metdata is only obtained from the backend when the model, emissions, or variable selection changes. This is the responsibility of the app controllers, way up the component tree. Metadata passes as props from there down the tree. 
- `props.dataToGraphSpec`, which mixes up a batch of special objects to please C3. This is based on 
  - results of `this.getMetadatas` above
  - `state.data` (which is gotten from the backend when and only when the metadata, area, or current dataSpec changes). No (extra) backend calls here either.

rod-glover: Hmm, I wonder if this explanation belongs in comments somewhere.

src/components/graphs/AnnualCycleGraph/AnnualCycleGraph.js
```diff
@@ -54,72 +54,148 @@ export default class AnnualCycleGraph extends React.Component {
...
+  componentDidUpdate(prevProps, prevState) {
+    if (!this.state.fetchingData && this.state.data === null) {
+      this.fetchData();
+    }
+  }
+
+  // Data fetching
+
+  getAndValidateData(metadata) {
+    const validateData = multiYearMeanSelected(this.props) ?
```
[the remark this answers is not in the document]
rod-glover: Hmm. I think this was copied from pre-refactoring code. 
Regardless, that's a good point, and this can be removed.

rod-glover:
Rebased

Issue #259, opened 2018-12-13
Update LongTermAveragesGraph lifecycle to match AnnualCycleGraph?


rod-glover:
ACG took some considerable thought to get right w.r.t. React lifecycle management.
Check LTAG to see whether the state change (selector: time of year) triggers a data fetch, and if so, use the pattern in ACG, which prevents trying to render with mismatched selector state and data state. This doesn't currently result in rank errors, but it is probably performing unnecessary updates and passing transiently (and without exploding) through an inconsistent state that it would be better to avoid. I suspect the brief double-update seen when changing Time Of Year is due to this problem, but have not verified with tracing yet.

Issue #261, opened 2018-12-14
Bug in export for StatisticalSummaryTable


rod-glover:
Call at line 155 to `exportDataToWorksheet` gets an argument from `this.state.dataTableTimeScale`, which no longer exists. This was likely dropped in fixing the `timeOfYear` bug, so consult the old code.

PR #262, opened 2018-12-14
Update LTAG and SST lifecycle to match AnnualCycleGraph

rod-glover:
Resolves https://github.com/pacificclimate/climate-explorer-frontend/issues/259

This PR eliminates the case where `state.timeOfYear` and `state.data` are transiently inconsistent. The equivalent situation produced a visible bug in ACG, but not here (by chance). This PR ports the learning from ACG to LTAG and SST.

In passing a [bug](https://github.com/pacificclimate/climate-explorer-frontend/issues/261) was noted in the SST export code.

Note to reviewers: This should be an easy one. It is just the same thing as a previous PR, repeated twice.


rod-glover:
[Demo](http://docker-dev02.pcic.uvic.ca:30507)

Issue #225, opened 2018-11-19
Flexible determination of "matching" climatologies


corviday:
We have a climatology period mismatch in our future climatologies:

| Downscaled Output & Climdex Indices | Degree Days & Return Periods |
|-----------------------------------|----------------------------|
| 2010-2039                         | 2011-2040                  |
| 2040-2069                         | 2041-2070                  |
| 2070-2099                         | 2071-2100                  |

Trevor Murdock indicates it is scientifically acceptable to compare these off-by-one-year climatologies in the comparison graphs like the Variable Response graph or the Annual Cycle graph. 

At present, when a user attempts to compare two datasets with off-by-one-year climatologies, the second dataset does not appear on  graphs, as no match can be found for the data in the first set. 

Make the process of locating matching datasets for comparison flexible enough to include these off-by-one climatologies.

PR #263, opened 2018-12-14
Flexible determination of "matching" climatologies (start and end dates)

rod-glover:
Resolves https://github.com/pacificclimate/climate-explorer-frontend/issues/225

This PR modifies the matching function used to select comparand climatologies matched to the primary climatology. It allows for start and end dates with slightly differing values, in this case by 1 year.


rod-glover:
Agree completely.

PR #268, opened 2019-01-03
Generate time-of-year object for data export

### LINKED ISSUE #261: Bug in export for StatisticalSummaryTable

corviday:
The file export function, `export.exportDataToWorksheet()`, requires a `selection` argument indicating what subset of data the user has selected for viewing. This corresponds to whatever the user has picked in the dropdown menu on the graph or table: a time-of-year `{time scale, time index}` object for the Statistical Summary Table and Long Term Average Graph, and an instance `{climatology period, run}` object for the Annual Cycle Graph. The information in the selection object is added to the filename and the explanatory metadata listed in the first part of the file, to indicate what's in the file.

At some point, refactoring broke the generation of the time-of-year selection object for the Statistical Summary Table. This PR fixes that.

The process of generating time-of-year selection objects was unnecessarily complicated by the fact that `exportDataToWorksheet` expected a time-of-year object with the attributes "timeres" and "timeidx" when the rest of CE constructs and uses time-of-year objects with the attributes "timescale" and "timeidx". So this PR also standardizes the timeres/timescale attribute name of `exportDataToWorksheet`'s time-of-year object for readibility and simplicity.

--------------------------------------------------------------------

It would be nice to figure out a way to write tests for `exportDataToWorksheet`. All its helper functions are well-tested, but I wasn't able to figure out how to catch and test the generated file. So bugs like this slip through our test suite. Maybe it would be a good idea to try and move even more of exportDataToWorksheet into testable helper functions.

### DIFF
--- a/src/components/DataControllerMixin/DataControllerMixin.js
+++ b/src/components/DataControllerMixin/DataControllerMixin.js
@@ -67,7 +67,7 @@ var ModalMixin = {
 
   exportDataTable: function (format) {
     exportDataToWorksheet("stats", this.props, this.state.statsData, format, 
-        {timeidx: this.state.dataTableTimeOfYear, timeres:this.state.dataTableTimeScale});
+        {timeidx: this.state.dataTableTimeOfYear, timescale:this.state.dataTableTimeScale});
   },
 
   injectRunIntoStats: function (data) {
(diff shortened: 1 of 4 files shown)


rod-glover:
Good catch. My bad. 

I agree re. testing, and your suggested approach is I believe the right one.

corviday:
I'm just glad you caught the bug, since it involved a totally untested function.

rod-glover:
Oh, this is a PR, not just a bug report! Let me know when you'd like me to review.

Also, is there an issue this PR should be linked to?

corviday:
It's for #261 , sorry. I had it in the title originally, but renamed it so it would be more obvious who whoever looks over merge messages later to update NEWS. :)

PR #270, opened 2019-01-07
Support substituting user-friendly units, resolve #169

### LINKED ISSUE #169: "kg m-2 day-1" is not a user friendly unit for precipitation
Both Trevor Murdock and MOTI engineers have indicated that while precipitation is often output as `kg m-2 d-1` from models, `mm/day` would be clearer to engineers, planners, and other non-climate-scientists.

One solution would be to add some sort of units attribute to the [variable configuration file](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/variable-options.yaml). Am advantage of this approach is that it would only apply to precipitation, which would be reasonable since measuring water is the only circumstances under which a kilogram per meter is equal to a millimeter.

[See also the back issue with precipitation units described here](https://github.com/pacificclimate/climate-explorer-backend/issues/83): that precip units are given by some files as `kg d-1 m-2` and others as `kg m-2 d-1`, which results in an error when trying to generate graphs with "different" units - which provides a reason to fix this issue on the back end instead.

corviday:
Addresses #169 

Adds a `userFriendlyUnits` attribute to the variable configuration file to be configured individually for each variable. This attribute's value is a list of pairs of unit strings (UNITSTRING1: UNITSTRING2) where UNITSTRING2 is an equivalent but more user-friendly version of UNITSTRING1. When displaying information from the backend API to the user, if the units attribute of the API results is UNITSTRING1, UNITSTRING2 will be substituted in graph labels, tables, and file exports instead.

For example:
```yaml
pr:
  userFriendlyUnits:
    - kg m-2 d-1: mm/day
    - kg d-1 m-2: mm/day
```

Would cause precipitation data received from the backend with units of kg m-2 d-1 to be labeled with units of mm/day. If some other variable was given in kg m-2 d-1, it would be unaffected by this attribute on pr. If precipitation was given in inches/day, it would also be unaffected.

This functionality affects only the units string. It does no numerical conversion and could not be used to convert `inches/day` to `mm/day` or similar.

### DIFF
--- a/src/core/util.js
+++ b/src/core/util.js
@@ -37,15 +37,18 @@ export function parseBootstrapTableData(data, metadata) {
   return _.map(data, function (stats, model) {
     const modelMetadata = _.find(metadata, m => m.unique_id === model);
     const period = `${modelMetadata.start_date} - ${modelMetadata.end_date}`;
+    const variable_id = modelMetadata.variable_id;
+    const configuredPrecision = getVariableOptions(variable_id, "decimalPrecision");
+    const precision = configuredPrecision ? configuredPrecision : PRECISION;
     const modelInfo = {
       model_period: period,
       run: stats.run,
-      min: +stats.min.toFixed(PRECISION),
-      max: +stats.max.toFixed(PRECISION),
-      mean: +stats.mean.toFixed(PRECISION),
-      median: +stats.median.toFixed(PRECISION),
-      stdev: +stats.stdev.toFixed(PRECISION),
-      units: stats.units,
+      min: +stats.min.toFixed(precision),
+      max: +stats.max.toFixed(precision),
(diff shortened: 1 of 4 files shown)


src/core/util.js
```diff
@@ -149,6 +151,29 @@ export function getVariableOptions(variable, option) {
...
+  const units = data.units;
+  const userFriendlyUnits = getVariableOptions(variable_id, "userFriendlyUnits");
+  if(userFriendlyUnits) {
+    for(let i = 0; i < userFriendlyUnits.length; i++) {
+      if(units in userFriendlyUnits[i]) {
+        return userFriendlyUnits[i][units];
+      }
+    }
+  }
+  return units;
```
rod-glover: You might be able to get underscore to do the work for you here.

Issue #267, opened 2018-12-14
Annual-only climatology graphs overlap exactly


rod-glover:
Possibly a minor concern.

To reproduce:

1. In Compare Variables portal, choose Var 1 = cdd; Var 2 = fdd (this will work for any pair of degree-day or return period datasets).
1. Graph shows a single line, which is actually two lines, one for each variable, that are plotted at exactly the same location, presumably because of how automatic axis scale is chosen for each.

This is slightly confusing, since data appears to be missing.

Issue #144, opened 2018-04-03
Reconsider matching time resolutions on extreme precipitation portal


corviday:
At present, Climate Explorer, when comparing two variables on the same graph, matches the resolution of the secondary variable to the primary variable. My thinking on this was that comparing variables would only make sense if they had the same time resolution.

However, I'm not sure this completely holds for the Extreme Precipitation portal. Many of the climdex values are annual-only, representing the number of days in a given year that whatever precipitation related thing happens. Upon playing with the annual climdices in the Extreme Precipitation portal, I think it would actually be more helpful to compare values of the form `[days in a year x precipitation thing happens]` against a timeseries showing monthly or seasonal precipitation (or even daily, but we don't have that) than against a single annual total precipitation value.

corviday:
Possibly add a prop to` DualDataController` that would control whether or not it limited secondary variable resolutions to resolutions available in the primary.

I still feel like matching resolutions is probably the right way to handle comparing non-climdex variables like temperature and precipitation.

corviday:
Summary: 
The underlying question here is: **when is it useful to compare two variables measured at different resolutions on a graph**? 

At present, climate explorer treats the answer as **"never"**, resulting in these flatline graphs where annual-resolution climdexes are only be compared with annual-resolution precipitation values, even when higher-resolution precipitation data is available.

jameshiebert:
@tqmurdock Do you have any opinions on this issue?

PR #272, opened 2019-01-08
Visually disambiguate annual-only MYM datasets on Annual Cycle Graph

corviday:
Addresses #267 
Annual-only multi-year means appear as horizontal lines on Annual Cycle Graphs - they have only a single value throughout the year. When graphing two such datasets at once, the lines are hard to see because C3 automatically scales each y-axis to frame data in the graph and draws the two horizontal lines on top of eachother, like this:
before-shift

This PR adds the `shiftAnnualCycle` attribute to all variables that are annual-only multi-year-mean datasets in the variable configuration file, signalling `DualAnnualCycleGraph` to pad the y axes of the graph to vertically translate the graph lines apart from eachother.

Variables changed: all the Degree Day and Return Period variables.

- cdd
- fdd
- gdd
- hdd
- rp5pr
- rp20pr
- rp50pr
- rp5tasmax
- rp20tasmax
- rp50tasmax
- rp5tasmin
- rp20tasmin
- rp50tasmin

With shifted annual cycles, the graph above looks like this:
after-shift

### DIFF
--- a/variable-options.yaml
+++ b/variable-options.yaml
@@ -156,45 +156,222 @@ wsdiETCCDI:
   decimalPrecision: 0
 
 #degree-day variables
+# these are annual-only multi year mean datasets, so they have a
+# graphable annual cycle, but it's a flat line.
+# apply shiftAnnualCycle to the other annual-only flat graph MYM
+# datasets.
 cdd:
   decimalPrecision: 0
+  shiftAnnualCycle:
+    - fdd
+    - gdd
+    - hdd
+    - rp5pr
+    - rp20pr
+    - rp50pr
+    - rp5tasmax
+    - rp20tasmax
+    - rp50tasmax
+    - rp5tasmin
+    - rp20tasmin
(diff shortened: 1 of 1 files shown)


corviday:
A thought: variable-based solution may be the wrong approach.

It's still possible to run into the two-horizontal lines displayed on top of eachother issue. DualAnnualCycleGraph matches available data resolutions in the secondary variable to the primary variable. So if, for example, you pick heating degree days as the primary variable (annual only) and compare it to tasmax (monthly, seasonal, and annual available), only annual tasmax will be displayed.

_Should_ this be how DualAnnualCycleGraph behaves? Can monthly data for one variable offer some insights to understanding a yearly variable? Good question; it's got its own issue, #144 

But the solution here probably should be "DualAnnualCycleGraph checks to see if all data is annual-resolution (either due to resolution matching, or availability) and pads the y axes if so".

corviday:
Closing issue to switch approaches.

PR #281, opened 2019-03-12
Update naming of CE

rod-glover:
Following Francis' requested changes:

About   CE

About   Credits

### DIFF
--- a/src/components/guidance-content/about/ClimateExplorer.js
+++ b/src/components/guidance-content/about/ClimateExplorer.js
@@ -7,7 +7,7 @@ export default function ClimateExplorer() {
     <Grid fluid>
       <Row>
         <FullWidthCol>
-          <h1>Climate Explorer</h1>
+          <h1>PCIC Climate Explorer (PCEX)</h1>
         </FullWidthCol>
       </Row>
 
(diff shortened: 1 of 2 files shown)


rod-glover:
@corviday , would you mind verifying these changes against these requested changes?

[PCEX_About_FZ_edits.pdf](https://github.com/pacificclimate/climate-explorer-frontend/files/2958930/PCEX_About_FZ_edits.pdf)

src/components/guidance-content/about/ClimateExplorer.js
```diff
@@ -17,11 +17,11 @@ export default function ClimateExplorer() {
...
               <p>
                 A tool for visualizing and downloading
-                climate model data and data derived from climate outputs
-                for the B.C. - Yukon region.
+                climate model outputs and data derived from those outputs.
               </p>
               <p>
-                Climate Explorer is also fondly known as 'the Marmot,'
+                The PCIC Climate Explorer (PCEX) is also fondly known as
+                the 'marmot,'
```
corviday: An unrequested (though trivial) change you've made:

`'the marmot'` 
to 
`the 'marmot'`

rod-glover: It was intentional; it's one of those trad vs. modern things. Trad says punctuation goes inside quotes. I'm kind of a trad guy, but I really don't care very much.

corviday: I  certainly agree with you on that! The difference I was commenting on is that "the" was previously inside the quotes and is now outside them.

PR #288, opened 2019-03-13
Improve docker instructions in README

### LINKED ISSUE #287: Improve docker instructions in README

rod-glover:
Resolve #287

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,19 +64,55 @@ If you *really* want to skip the linting during a commit, you can always run `gi
 ### Setup using Docker
 
 
-#### Build docker image
+#### Build docker image manually
 
-Clone repo:
+Until recently (roughly, Jan 2019), we were using Dockerhub automated builds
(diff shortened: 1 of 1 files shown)


corviday:
The content seems accurate and clear. Do these site-specific details make more sense in Confluence than in the public README, though?

rod-glover:
That's a good question. I wanted to avoid having to go to multiple places to get a complete description of how to build and deploy manually. I still want that; I suppose we could migrate this to a specific Confluence page that is linked in the README. Question: Is it worth the extra effort and complication?

corviday:
>  _Is it worth the extra effort and complication?_

I lean toward "no". There's already some site-specific information in the README, and I doubt anyone not associated with PCIC is ever going to run this, so where the information goes is a matter of aesthetics, more than utility.

rod-glover:
Good, I lean that way too. And it's not as if we can't change our minds easily.

PR #290, opened 2019-03-27
Improve UI for degree days. Resolve #289, resolve #277

### LINKED ISSUE #277: Reconsider the use of the word "Mean" in data series labels
Currently, each data series on a chart is labeled with a dynamically generated series of attributes drawn from the dataset metadata, followed by the word "Mean." Such as: 
- Yearly Mean
- Tasmax Mean
- Yearly Tasmax Mean
- Tasmax 2010-2039 Mean

The word "Mean" is hard coded; everything else is derived from comparing the metadata for each dataseries and including each attribute (resolution, variable, period, etc) which varies between timeseries in the legend.

However, I think there are some variables, specifically return period datasets and degree day datasets, where the hardcoded "Mean" might not be the clearest possible phrasing.. It's not technically inaccurate - these data series still represent climatological means - but it seems less obvious that this is the right phrasing than for datasets like tasmax or tasmin, which represent the climatological means of monthly or seasonal means of a daily dataset. 

I'd like to ask Trevor if he has any better ideas for dataset labels, and if so, implement them. He may not, though, which isn't a big deal - detailed descriptions of what variables represent is available in the help. It may not be possible to improve on the short summary in the legends.

Low priority.

### LINKED ISSUE #289: Add note that annual degree days is a sum, not a mean to help
Unlike every other data type in PCEX, the "annual" value of degree_days is the _sum_ of the seasonal / monthly values, not their _mean_. This fact deserves a sentience in the degree day data description.

corviday:
Makes two minor UI changes intended to clarify the meaning of degree day variables:

1. Adds a new configurable variable display option named `seriesLegendString`. This option affects the caption that will be given as the name of each data series in the Annual Cycle, Change From Baseline, and Timeseries graphs. 
Previously, because these graphs display multi year means, all legends on these graphs were hardcoded to use the word "Mean" as a base name, such as "Tasmax Mean", "r1i1p1 Mean", or "Tasmax r1i1p1 Mean". With this change, the base name of a data series description can be set in the variable configuration file, but if no base name is set, the default is still "Mean." The basename for frost days and all degree day variables has been set to "Mean Count." All other variables default to "Mean."

2. Adds a sentence to the description of degree days in the Help explicitly noting that if you look at a graph for degree days, the annual values will be larger than the seasonal values (since annual values are the sum, not the mean, of seasonal values).

We haven't received any complaints that degree days were confusing, but they do behave differently from some other data types, and I've made some mistakes, so I wanted the UI to clarify their workings a bit.

Addresses #289 
Addresses #277

### DIFF
--- a/src/core/chart-generators.js
+++ b/src/core/chart-generators.js
@@ -385,7 +385,9 @@ function getMonthlyData(data, timescale = 'monthly') {
  * the graph legend will need to display variable names as well.
  *
  * Timeseries names include any descriptive  metadata that vary between
- * timeseries and leave out any metadata that doesn't. They end with "mean".
+ * timeseries and leave out any metadata that doesn't. They end with a
+ * basename specific to the variable and set in the variable config file 
+ * (or "mean" if the config file doesn't specify).
  */
 // TODO: special case climatological period to display as (XXXX-XXXX)
 // TODO: possibly cue descriptors to appear in a specific order?
@@ -425,13 +427,23 @@ function shortestUniqueTimeseriesNamingFunction(metadata, data) {
   if (variation.length === 0) {
     throw new Error('Error: cannot graph identical timeseries');
   }
+  
+  // Build a dictionary with the base name for each variable (typically
+  // either "Mean" or "Mean Count") from the variable config file. 
+  // Defaults to "Mean" since all data displayed by this graph is MYMs.
+  const variables = _.uniq(_.pluck(metadata, "variable_id"));
+  function getVarBasename(v) {
+    const fromConfig = getVariableOptions(v, 'seriesLegendString');
+    return _.isUndefined(fromConfig) ? "Mean" : fromConfig;
+  }
+  const basenameByVariable = _.object(variables, _.map(variables, getVarBasename));
 
   return function (m) {
     let name = '';
     for (let v of variation) {
       name = name.concat(`${m[v]} `);
     }
-    name = name.concat('mean');
+    name = name.concat(basenameByVariable[m.variable_id]);
     return capitalizeWords(name);
   };
 }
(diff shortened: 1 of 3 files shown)


corviday:
Don't merge until #308 is done.

corviday:
#308 is done, I think this is ready to go!

src/core/chart-generators.js
```diff
@@ -425,13 +427,23 @@ function shortestUniqueTimeseriesNamingFunction(metadata, data) {
   if (variation.length === 0) {
     throw new Error('Error: cannot graph identical timeseries');
   }
+  
+  // Build a dictionary with the base name for each variable (typically
+  // either "Mean" or "Mean Count") from the variable config file. 
+  // Defaults to "Mean" since all data displayed by this graph is MYMs.
+  const variables = _.uniq(_.pluck(metadata, "variable_id"));
```
rod-glover: Underscore `_.pluck` is replaced in Lodash by `_.map` (special case of a string iteratee). One or the other of us will have to make this change, depending on in what order we merge PRs (mine is #317). How do you want to do this?

src/core/chart-generators.js
```diff
@@ -425,13 +427,23 @@ function shortestUniqueTimeseriesNamingFunction(metadata, data) {
...
+  function getVarBasename(v) {
+    const fromConfig = getVariableOptions(v, 'seriesLegendString');
+    return _.isUndefined(fromConfig) ? "Mean" : fromConfig;
+  }
+  const basenameByVariable = _.object(variables, _.map(variables, getVarBasename));
```
rod-glover: Similarly, `_.object` has to be replaced by something different, this time a compound of Lodash `_.fromPairs` and `_.zipObject`. (See [Migrating](https://github.com/lodash/lodash/wiki/Migrating).) Again, timing.

rod-glover:
Having thought this over for another couple of minutes, I'd prefer to merge my PR first, then help you fix the brokenness in this one. It will be easy to do, and I am uneasy about rebasing my migration PR, even just a little.

corviday:
Sure, go ahead. This PR has already had to be rebased when master was refactored twice, a third time won't hurt anything. :)

rod-glover:
Mine is merged. You are good to rebase.

corviday:
This PR has been updated for lodash, and should be ready to go.

Issue #216, opened 2018-10-17
It's hard to find the variable you want in the long list of variables


rod-glover:
When the variables available include Climdex variables, degree-day variables, and/or return period bariables, the list is long. It's hard to find the variable you are interested in.

Options:

1. Sort list alphabetically.
1. Offer a filter tool to find the option in the list that you want. Betcha there's already such a selector out there in OSS!

rod-glover:
The more I use CE, the more I dislike how the current variable selector works. Suggest adopting a selector component with the following criteria (ranked by importance):

Imp | Criterion | [React Select](https://jedwatson.github.io/react-select/) | Candidate B | Candidate C
-|-|-|-|-
1 | Mature, maintained | ✔️ | 
1 | Dropdown | ✔️ | 
1 | Search/filter | ✔️ | 
1 | Disable selector | ✔️ | 
1 | Disable options | ✔️ | 
1 | Flexible, configurable | ✔️ | 
1 | Aesthetics | ✔️ | 
2 | Return arbitrary value | ✔️ | 
2 | Multiselect |  ✔️ |

rod-glover:
Merry Christmas 🎄

rod-glover:
Resolved with the new selector, a long time ago

PR #310, opened 2019-07-09
Address #309, arrange variables in dropdown menu

### LINKED ISSUE #309: Add a way to control the order variables appear in the dropdown
Closely related: #216 

Variables are displayed in the variable selector dropdown in the order they first appear in a dataset returned by the `multimeta` query. The `multimeta` query displays datasets in the order they appear in in the `data_files` table, which is, in turn, the order they were added to the database in.

I have recently updated some datasets in the database and in the PCEX ensemble. Some non-climatology datasets have been removed from the ensemble and replaced with climatological version. However, since I just created the climatological versions, they are at the end of `data_files` and appear at the end of `multimeta`.

This has had the mildly unfortunate side effect of causing the variables of the updated climdex datasets to appear at the bottom of the variable selection menu, separated from all the other climdex variables. Since variables are described and defined by "type" in the help, it was nice that all variables of a "type" (model output, climdex, return period, degree day) appeared together in the drop down menu. It made it easier to find related measures, and easier to find information on a variable in the Help.

It looks like #216 will fix this eventually. Adding some sort of variable type indicator to the variable configuration file and using that to group related variables together when constructing the menu would be a quick fix, if needed.

corviday:
Currently, the order variables appear in the variable selection dropdown menu matches the order the first instance of each variable appears in the multimeta query. Historically, this has worked adequately for us, because we have added data in stages. We started with model output, then added climdex data, then degree day data, and finally return period data. 

The multimeta query returns datasets in no particular order, but it approximates the order they were added to the database well enough, so historically variables have appeared in the menu with the related variables that were added at the same time. All the model outputs together, all the degree days together, etc.

Unfortunately, I just updated a bunch of data ( https://github.com/pacificclimate/climate-explorer-data-prep/issues/73 ) and now the order modelmeta is returning data in is a lot less convenient and the default is a much more confusing menu. Many of the updated climdex variables now appear at the end of the variable dropdown, far separated from the rest of the climdex variables, and much harder to find (or look up in the help).

This PR orders the variables inside the variable selection dropdown. Variables are assigned into numerical "menu groups" via the new `menuGroup` attribute in the variable configuration file. The menu groups could be anything, but in this case, they're set to correspond to the groupings used in the data documentation:

0. model output data
1. climdex data
2. degree day data
3. return period data

Variables not assigned to any group by the variable configuration file are assigned to group `MAX_SAFE_INTEGER` and appear at the end of the menu. With each group, variables are listed in alphabetical order.

This is something of a temporary patch; long term variable selection is expected to work completely differently ( https://github.com/pacificclimate/climate-explorer-frontend/issues/216 ) . However, I think short term, making it easier to look up variables in the help file and find related variables is worth this quick fix.

Resolves #309

### DIFF
--- a/src/components/VariableDescriptionSelector/VariableDescriptionSelector.js
+++ b/src/components/VariableDescriptionSelector/VariableDescriptionSelector.js
@@ -26,6 +26,7 @@ import _ from 'underscore';
 import IndexedSelector from '../IndexedSelector';
 import React from 'react';
 import PropTypes from 'prop-types';
+import { getVariableOptions } from '../../core/util';
 
 export default class VariableDescriptionSelector extends React.Component {
   static propTypes = {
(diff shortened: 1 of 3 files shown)


src/components/VariableDescriptionSelector/VariableDescriptionSelector.js
```diff
@@ -42,10 +43,21 @@ export default class VariableDescriptionSelector extends React.Component {
...
+	// variables are sorted into groups based on the "menuGroup"
+	// attribute in the variable configuration file.
+	// Within a group, they're alphabetical by variable_id
+	function sortVariables(variables) {
+	  let groups = _.groupBy(variables, v => {
+		const membership = getVariableOptions(v.variable_id, "menuGroup");
+		return _.isUndefined(membership) ? Number.MAX_SAFE_INTEGER : membership;});
+	  groups = _.map(groups, g=> {return _.sortBy(g, 'variable_id');});
+	  return _.flatten(groups);
+    }
```
rod-glover: Nice.

corviday: Underscore is pretty great. :)

PR #313, opened 2019-07-11
Make 2.x deployable for production

### LINKED ISSUE #311: Make 2.x versions deployable in production environment

### LINKED ISSUE #312: Inject configuration values at runtime (workaround for absent runtime env variables)
CRA solves a lot of problems but it introduces one for us too: (RTFM) It injects environment variables only at _build time_, not at run time. ["The environment variables are embedded during the build time. Since Create React App produces a static HTML/CSS/JS bundle, it can’t possibly read them at runtime."](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables)

CRA [suggests](https://facebook.github.io/create-react-app/docs/title-and-meta-tags#injecting-data-from-the-server-into-the-page) a workaround that requires adding extra tooling in the server that substitutes values from the runtime environment into the HTML that is served, and these values are then hoisted into the app. Inelegant, adds quite a bit of complexity to the app infrastructure.

We already have a solution in place in our codebase: We already retrieve configuration information from the `public` folder, e.g., for variable options. We can do the same thing for the values that we want to inject using environment variables, only now they won't be environment variables, they will be a (YAML) configuration file that we mount as a volume and can modify dynamically. You might not even have to restart the app.

I think it would be best, if possible, to make this appear to be run-time environment variables, meaning that the values provided should be used to modify `process.env[<variable name>]`. This would minimize changes to the existing code, and it _might_ address the problem of setting PUBLIC_URL, which is very important to being able to deploy the same build (and therefore the same docker image) with an arbitrary base URL.

rod-glover:
Resolves #311 
Resolves #312

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,7 @@ COPY package.json /app/package.json
 RUN npm install --quiet
 RUN npm install -g serve
 COPY . /app
-RUN npm run build
 
 EXPOSE 8080
 
-CMD serve -s build -l 8080
+CMD ["/bin/bash", "./entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # This Dockerfile adapted from https://mherman.org/blog/dockerizing-a-react-app/
-# There are a *lot* of approaches to this task. This one does work. Yay us.
 
-FROM node:latest
+# At this moment, Node.js 10.16 LTS is recommended for most users.
+FROM node:10.16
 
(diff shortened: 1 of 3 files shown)


rod-glover:
A test deployment of this branch is running at https://services.pacificclimate.org/dev/pcex/app/

Note the problematic double-`/ `in the routing was due to an incorrect configuration (env. var), now corrected in the `fe.env` file consumed by `docker-compose`.

PR #317, opened 2019-07-13
Migrate underscore to lodash

### LINKED ISSUE #252: Migrate underscore to lodash
* It's a superset. In particular, it's got the extremely valuable `_.get` and related functions, which we really, really like. (An Underscore plugin called `undescore.get` is buggy. Yeesh.)
* Its API is more consistently defined, named and argumented.
* It includes [highly FP-friendly versions](https://github.com/lodash/lodash/wiki/FP-Guide) of the functions.
* It's widely regarded as better implemented than underscore.
* I've used it in other projects with much happiness.
* I'm using it in projects we'd like to use in CE, e.g., `pcic-react-components` (which needs to be renamed `pcic-react-selectors`, but that's another matter).
* [There are migration tools](https://github.com/lodash/lodash/wiki/Migrating).

What's not to like?
https://lodash.com/docs

rod-glover:
Resolves #252 

Q: Why?
A1: Packages we want to use depend on it (peer dependency, or ought to be).
A2: [Underscore is not being maintained and is now regarded as deprecated](https://blogs.dropbox.com/tech/2018/09/migrating-from-underscore-to-lodash/).

From [Lodash wiki: Migrating](https://github.com/lodash/lodash/wiki/Migrating):

- [x] Underscore _.any is Lodash _.some
- [x] Underscore _.all is Lodash _.every
- [x] Underscore _.compose is Lodash _.flowRight
- [x] Underscore _.contains is Lodash _.includes
- [x] Underscore _.each doesn’t allow exiting by returning false
- [x] Underscore _.findWhere is Lodash _.find
- [x] Underscore _.flatten is deep by default while Lodash is shallow
- [x] Underscore _.indexOf with 3rd parameter undefined is Lodash _.indexOf
- [x] Underscore _.indexOf with 3rd parameter true is Lodash _.sortedIndexOf
- [x] Underscore _.indexBy is Lodash _.keyBy
- [x] Underscore _.invoke is Lodash _.invokeMap
- [x] Underscore _.mapObject is Lodash _.mapValues
- [x] Underscore _.max combines Lodash _.max & _.maxBy
- [x] Underscore _.min combines Lodash _.min & _.minBy
- [x] Underscore _.sample combines Lodash _.sample & _.sampleSize
- [x] Underscore _.object combines Lodash _.fromPairs and _.zipObject
- [x] Underscore _.omit by a predicate is Lodash _.omitBy
- [x] Underscore _.pairs is Lodash _.toPairs
- [x] Underscore _.pick by a predicate is Lodash _.pickBy
- [x] Underscore _.pluck is Lodash _.map
- [x] Underscore _.uniq by an iteratee is Lodash _.uniqBy
- [x] Underscore _.where is Lodash _.filter
- [x] Underscore _.isFinite doesn’t align with Number.isFinite
      (e.g. _.isFinite('1') returns true in Underscore but false in Lodash)
- [x] Underscore _.matches shorthand doesn’t support deep comparisons
      (e.g. _.filter(objects, { 'a': { 'b': 'c' } }))
- [x] Underscore ≥ 1.7 & Lodash _.template syntax is
      _.template(string, option)(data)
- [x] Lodash _.memoize caches are Map like objects
- [ ] Lodash doesn’t support a context argument for many methods in favor of _.bind
- [x] Lodash supports implicit chaining, lazy chaining, & shortcut fusion
- [x] Lodash split its overloaded _.head, _.last, _.rest, & _.initial out into
      _.take, _.takeRight, _.drop, & _.dropRight 
      (i.e. _.head(array, 2) in Underscore is _.take(array, 2) in Lodash)
    - [x] _.head
    - [x] _.last
    - [x] _.rest
    - [x] _.initial

Additionally:

- [x] Underscore _.uniq with a boolean 2nd param is Lodash _.sortedUniq
- [x] Underscore _.uniq with a boolean 2nd param and iteratee 3rd param is Lodash _.sortedUniqBy

Found during testing:

- [x] Underscore _.min, _.max return min, max of values in an object; Lodash _.min, _.max returns undefined for an object (e.g., Underscore _.min({ a: 2, b: 1}) === 1; Lodash _.min({ a: 2, b: 1}) === undefined)
- [x] Underscore _.min, _.max (apparently) ignore non-number members of the argument; Lodash does not; behaviour is complicated when non-numbers are present, but it will sometimes return a string value (even if it does not evaluate to a number).
- [x] Underscore _.matcher is Lodash _.matches

### DIFF
--- a/src/core/__tests__/chart-accessor-tests.js
+++ b/src/core/__tests__/chart-accessor-tests.js
@@ -71,7 +71,6 @@ describe('yAxisRange', function () {
   const monthlyTasmaxTimeseriesData = _.map(monthlyTasmaxTimeseries.data);
   const monthlyPrTimeseriesData = _.map(monthlyPrTimeseries.data);
   it('calculates the min and max of data associated with a y-axis', function () {
-    console.log('### monthlyTasmaxTimeseriesData', monthlyTasmaxTimeseriesData)
     expect(yAxisRange(graph, 'y').min).toBe(_.min(monthlyTasmaxTimeseriesData));
     expect(yAxisRange(graph, 'y').max).toBe(_.max(monthlyTasmaxTimeseriesData));
     expect(yAxisRange(graph, 'y2').min).toBe(_.min(monthlyPrTimeseriesData));
--- a/src/core/__tests__/chart-accessor-tests.js
+++ b/src/core/__tests__/chart-accessor-tests.js
@@ -68,10 +68,13 @@ describe('yAxisRange', function () {
   const graph = timeseriesToAnnualCycleGraph(metadataToArray(),
       monthlyTasmaxTimeseries,
       monthlyPrTimeseries);
+  const monthlyTasmaxTimeseriesData = _.map(monthlyTasmaxTimeseries.data);
+  const monthlyPrTimeseriesData = _.map(monthlyPrTimeseries.data);
   it('calculates the min and max of data associated with a y-axis', function () {
-    expect(yAxisRange(graph, 'y').min).toBe(_.min(monthlyTasmaxTimeseries.data));
-    expect(yAxisRange(graph, 'y').max).toBe(_.max(monthlyTasmaxTimeseries.data));
-    expect(yAxisRange(graph, 'y2').min).toBe(_.min(monthlyPrTimeseries.data));
-    expect(yAxisRange(graph, 'y2').max).toBe(_.max(monthlyPrTimeseries.data));
+    console.log('### monthlyTasmaxTimeseriesData', monthlyTasmaxTimeseriesData)
+    expect(yAxisRange(graph, 'y').min).toBe(_.min(monthlyTasmaxTimeseriesData));
+    expect(yAxisRange(graph, 'y').max).toBe(_.max(monthlyTasmaxTimeseriesData));
+    expect(yAxisRange(graph, 'y2').min).toBe(_.min(monthlyPrTimeseriesData));
+    expect(yAxisRange(graph, 'y2').max).toBe(_.max(monthlyPrTimeseriesData));
   });
 });
--- a/src/core/__tests__/chart-accessor-tests.js
+++ b/src/core/__tests__/chart-accessor-tests.js
@@ -18,10 +18,10 @@ import {monthlyTasmaxTimeseries,
         annualTasmaxTimeseries,
         monthlyPrTimeseries,
         metadataToArray} from '../__test_data__/sample-API-results';
-import _ from 'underscore';
+import _ from 'lodash';
 
 jest.dontMock('../chart-accessors');
-jest.dontMock('underscore');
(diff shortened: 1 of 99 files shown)


rod-glover:
[Demo](https://services.pacificclimate.org/dev/pcex/app)

rod-glover:
@corviday , would you be so kind as to exercise this thing? I have done so and find no problems, but I'd like your eyes on this too.

rod-glover:
TODO: Perhaps: Remove `_.min` and `_.max` debug mixins that check arg types. Possibly these should be left in for a while longer -- until the next release, say -- so that any issues arising from the rather different behaviour of Underscore vs. Lodash on these with arguments that are not pure arrays of numbers can be detected.

corviday:
I haven't the faintest idea why this would be, but when I run this branch and go to the Compare Variables portal, most of the variables are missing from the Variable 2 dropdown. Are you seeing that at all?

When I run current master, the variables are present.

src/core/__tests__/chart-accessor-tests.js
```diff
@@ -68,10 +68,13 @@ describe('yAxisRange', function () {
...
       monthlyTasmaxTimeseries,
       monthlyPrTimeseries);
+  const monthlyTasmaxTimeseriesData = _.map(monthlyTasmaxTimeseries.data);
+  const monthlyPrTimeseriesData = _.map(monthlyPrTimeseries.data);
   it('calculates the min and max of data associated with a y-axis', function () {
-    expect(yAxisRange(graph, 'y').min).toBe(_.min(monthlyTasmaxTimeseries.data));
-    expect(yAxisRange(graph, 'y').max).toBe(_.max(monthlyTasmaxTimeseries.data));
-    expect(yAxisRange(graph, 'y2').min).toBe(_.min(monthlyPrTimeseries.data));
-    expect(yAxisRange(graph, 'y2').max).toBe(_.max(monthlyPrTimeseries.data));
+    console.log('### monthlyTasmaxTimeseriesData', monthlyTasmaxTimeseriesData)
```
corviday: Possibly remove this debug output if you're feeling extra-tidy.

src/core/chart-accessors.js
```diff
@@ -6,7 +6,13 @@
  *
  * Reference on the C3 chart spec format can be found at https://c3js.org
  ***************************************************************************/
-import _ from 'underscore';
+import _ from 'lodash';
+
+
+// C3 data series are an array containing a string axis label followed by the
+// (numeric) data values for the axis. This function returns the data.
+export const seriesData = series => _.drop(series, 1);
```
corviday: Nice!

rod-glover:
@corviday , I made a mistake in the migration. Fixed it. Looks OK now. Do you concur?

[Demo](https://services.pacificclimate.org/dev/pcex/app)

PR #323, opened 2019-07-18
Resolve #322; update default model

### LINKED ISSUE #322: Make PCIC12 the default "model"
Now that we have ensemble averages available, we should use them by default. Most people will not care about individual models (and using them without knowing what you're doing is inadvisable). Let's set the defaults to PCIC12 and RCP8.5.

corviday:
Updates the default model to `PCIC12` for all portals. 

(The Extreme Precipitation portal doesn't actually have access to any PCIC12 data, so it falls back onto an arbitrary model, which seems, in practice, to be CanESM2.)


rod-glover:
LGTM

PR #331, opened 2019-08-16
Use pcic-react-external-text package

### LINKED ISSUE #304: Extract external texts module as a Node package
This package will be useful across many if not all of our React projects. In particular, a similar (but less featureful) module, which was the basis for this one, is in use in the Plan2Adapt v2 project.

rod-glover:
Resolves #304 : Extract external texts util to a package


rod-glover:
@corviday , this is very simple. The change line count is almost all in `package-lock.json`, which you can ignore.

Issue #334, opened 2019-08-23
Change From Baseline throws error on _.max()


rod-glover:
`Error: _.max(): argument contains non-numbers`

See also https://github.com/pacificclimate/climate-explorer-frontend/issues/326, https://github.com/pacificclimate/climate-explorer-frontend/issues/324

PR #333, opened 2019-08-20
Use pcic-react-components selectors

rod-glover:
At long last we substitute the pcic-react-components selectors for the current ones. 

See https://github.com/pacificclimate/climate-explorer-frontend/issues/332. Some particularly gnarly selectors have not been replaced in this PR.
Fixes https://github.com/pacificclimate/climate-explorer-frontend/issues/325
Fixes #339 
Fixes https://github.com/pacificclimate/climate-explorer-frontend/issues/334 (incidentally)

This gives us the more featureful selectors and, flowing from the simplification of selectors' client code (app controllers, etc.), substantial additional improvements in the CE code, including:

- logically simpler code
- remove need for AppMixin (React recommends strongly against mixins)
- replace outdated React.createClass usage with standard ES6 class declaration
- use React recommended best practice for managing asyn data fetching (metadata)
- use HOCs to factor out some common, and somewhat complicated, code shared between app controllers

Which is all like, Yay!

What's not so Yay is that there are quite a few changes. The real change count (discounting package-lock.json) is about +762, -1035, but that's still quite a bit. Here's a guide to the substantive changes:

- DRY up and simplify app controllers, using the following. Despite the fact that there is a slightly higher line count, these components are now logically much simpler. Given the comprehensive changes to them, it's best not to look at the diffs, just at the changed code.

  - add HOC `withAsyncData`: Extracts the React best practice for managing async data fetching. It is applied to all app controllers now and can be applied in at least a couple of other places (in another PR). This may be moved to pcic-react-components.

  - Replace model, emissions scenario, and variable selectors with components from pcic-react-components. These components internally handle constraining (disabling) options based on selections in a cascade of selectors. This has enabled a lot of confusing code in app controllers to be removed.

  - Extract remaining common code in app controllers to app-controllers/common.js.

- Replace TimeOfYearSelector (in graphs and tables only; there's another similar one in MapSettings that needs replacement, but later...) with a new component (same name) in pcic-react-components.

- Make some components more robust to undefined values for props, which can happen now while selectors are doing cascading defaults.

- Fix a couple of subtle bugs related to mishandling asynchronous events.

- Add a lot of comments to new and existing code.

PR #337, opened 2019-08-23
Fix Lodash max error in Change From Baseline

rod-glover:
Fixes https://github.com/pacificclimate/climate-explorer-frontend/issues/334


rod-glover:
Handled in a [different PR](https://github.com/pacificclimate/climate-explorer-frontend/pull/333).

PR #341, opened 2019-08-29
Improve deploy version setting mechanism

### LINKED ISSUE #335: Need a more reliable mechanism for setting version info in deploys
The currently deployed CE shows a version of "2.0.2", and no other information. We don't have anything tagged 2.0.2, and without a commit or other unambiguous indicator, we have no way to determine what is actually running in this deploy.

Unfortunately the new deployment process makes it hard to automatically include the commit in the versioning, and so we get human error like this one. It would be best to work out an automatic process, but this will take effort.

rod-glover:
Resolves #335 

This PR allows the user to inject, at image build time, a mechanically generated version code into the Nodejs build process that happens when the image is run.

README provides details of how to do this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@
 # In future, as we scale up, we may want to use an Alpine base image, which would reduce
 # the size of the image by about an order of magnitude and reduce the attack surface of
 # the image as well.
+
+# docker build --build-arg REACT_APP_CE_CURRENT_VERSION=$(./generate-commitish.sh)
 FROM node:10.16
 
 ADD . /app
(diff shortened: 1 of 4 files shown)


rod-glover:
[Demo](http://docker-dev01.pcic.uvic.ca:30502/pcex/app/#/about/climate-explorer)

rod-glover:
@matthewbenstead would you mind reviewing this PR? Let's discuss it's implications for your deploy scripts.

matthewbenstead:
@rod-glover Sure - looking at the changes the only deployment change I can see is modifying the environment variables passed to the front end at run time. 

We're supplying the version here - by removing this line we should be able to use the default version that the generate-commitish.sh script add. 
```bash
# grep "REACT_APP_CE_CURRENT_VERSION" fe.env
REACT_APP_CE_CURRENT_VERSION=2.0.2
```

Now this should supply the build process with the "correct" version of the REACT_APP_CE_CURRENT_VERSION variable - does that variable get persisted somewhere in the image during the build process?

rod-glover:
Yes, by removing that line in `fe.env`, you will not be overriding the version that got built in to the Docker image, which is what we want. And yes, exactly, that variable gets persisted in the image. The mechanism to inject it is slightly tricky, but it's a standard thing for this kind of problem.

rod-glover:
Also, yes, that is what I thought you would say, but I wanted your independent and more expert eye on it. Thanks.

rod-glover:
After discussion with @matthewbenstead , we have determined that this PR is good to merge.

PR #346, opened 2019-09-25
Add Jenkinsfile

nikola-rados:
In order to run a pipeline on Jenkins a repo requires a `Jenkinsfile`.  This file is responsible for configuring the stages of a pipeline and the actions it takes.  The `Jenkinsfile` I have created simply runs our unit tests and then creates a docker image.

I added a new `npm run` command for running tests in a CI environment.  The regular testing command required user input to run through the tests which was causing hangups in Jenkins.

The image is named `climate-explorer-frontend` for pushes to master and `climate-explorer-frontend/[branch_name]` for all other branches.  I have not found any strict naming conventions for docker images, so this is open for us to discuss.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "start:rg": "CE_BACKEND_URL=https://services.pacificclimate.org/marmot/api NCWMS_URL=https://services.pacificclimate.org/marmot/ncwms TILECACHE_URL=https://tiles.pacificclimate.org/tilecache/tilecache.py CE_ENSEMBLE_NAME=all_downscale_files CE_CURRENT_VERSION=$(./generate-commitish.sh) react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!(leaflet|pcic-react-components)/)\"",
-    "ci": "CI=true react-scripts test --env=jsdom --transformIgnorePatterns \"node_modules/(?!(leaflet|pcic-react-components)/)\"",
+    "jenkins-test": "CI=true react-scripts test --env=jsdom --transformIgnorePatterns \"node_modules/(?!(leaflet|pcic-react-components)/)\"",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
     "start:rg": "CE_BACKEND_URL=https://services.pacificclimate.org/marmot/api NCWMS_URL=https://services.pacificclimate.org/marmot/ncwms TILECACHE_URL=https://tiles.pacificclimate.org/tilecache/tilecache.py CE_ENSEMBLE_NAME=all_downscale_files CE_CURRENT_VERSION=$(./generate-commitish.sh) react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!(leaflet|pcic-react-components)/)\"",
+    "ci": "CI=true react-scripts test --env=jsdom --transformIgnorePatterns \"node_modules/(?!(leaflet|pcic-react-components)/)\"",
```
corviday: This is a mildly confusing name, since `npm ci` does not actually invoke this command. I know the command isn't for human use, so this bit of confusion is unlikely to ruin anyone's day, but if there's a less confusing thing you could name it, that might be nice.

corviday:
I made a [branch](https://github.com/pacificclimate/climate-explorer-frontend/tree/test-jenkins) off this one to see if jenkins would build an image for it. As far as I can tell, it didn't, but I'm not sure if jenkins just isn't set up to do that yet, or I'm looking in the wrong place for the image, or there's something not working quite right with this branch.

nikola-rados:
@corviday the branch did in fact make it to jenkins but it failed during the testing stage.  I'll look into what the issue is.

nikola-rados:
I think this should be ready to go @corviday if you want to take one more look to check it off!

corviday:
Ignore me self-requesting a review. I was wondering what a button I'd never seen before did.

PR #349, opened 2019-11-13
Update Jenkins Pipeline

nikola-rados:
Here are the changes in this PR:
- automatically push images to [docker hub](https://hub.docker.com/orgs/pcic/repositories) using jenkins user
- updated naming to accommodate docker hub (ex. `pcic/[tool_name]:[branch_name]_[build_id]`)
- conduct security scan of image
- automate cleanup docker image on `dev01`

A concern I have with this change is the amount of tags that could get pushed to docker hub.  The `build_id` tag suffix refers to the number of pipeline builds which correlates to the number of pushes on that branch.  On the one hand, these unique tags could prove useful if a particular push breaks something, on the other hand, it may be overkill to create so many unique tags.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,8 +17,19 @@ node {
     String name = BASE_REGISTRY + 'climate-explorer-frontend'
 
     // tag branch
-    if (BRANCH_NAME != 'master') {
+    if (BRANCH_NAME == 'master') {
+        // figure out how to tag
+    } else {
         name = name + ':' + BRANCH_NAME + "_${BUILD_ID}"
+    }
+
+    stage('testing for payload string') {
+        def payloadString = build.buildVariableResolver.resolve("payload")
+        sh "${payloadString}"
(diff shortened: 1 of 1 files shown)


rod-glover:
Are tags costly? I would guess they are cheap, and that the only concern might be if there are so many that it is hard to wade through them in search of a particular one. But I doubt that is any more the case than with commits, which are numerous but not a problem because of that.

nikola-rados:
@rod-glover I don't think tags are costly, I just wasn't sure if the clutter on docker hub would be annoying for one reason or another.  As for the syntax, I'm using the _scripted_ jenkinsfile syntax, there is an alternative called _declarative_ that you may like better!  As far as I understand it there is some back and forth about which is "better", but they are fairly interchangeable.

rod-glover:
Other syntax complaints aside, in general declarative syntax is easier to understand and maintain. Your call natch, but if you want my opinion, declarative is preferred unless it offers hopelessly inferior functionality.

nikola-rados:
@jameshiebert that seems like a good change.

PR #351, opened 2019-12-04
Jenkins Tag Tracking

nikola-rados:
This PR introduces new code in the `Jenkinsfile` that handles git tags.  

### Getting Git Tags
The `get_tags` function checks the most recent commit for a tag and uses it to tag the image on docker hub.  Furthermore, if a tag is detected a `latest` tag is also pushed to docker hub.  A created tag and `latest` tag are only used for `master`.  If the branch is not `master` or `master` is not tagged it uses the branch name as the tag.

### Pipeline Duration
This iteration of the pipeline takes about 12 minutes to run.  Over half of that time is the `Security Scan` stage.  I thought I would open up some conversation about how often we want to run this stage since it doubles the pipeline duration.  I had an idea that we could run this stage once the branch is filed as a PR.  That way before merging Jenkins will still check for vulnerabilities but we can speed up development time.  Let me know what you think!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,7 +54,8 @@ def build_and_publish() {
 
     // if (BRANCH_NAME == 'master') {
     if (BRANCH_NAME == 'jenkins-tag-tracking') {
-        image_names.plus(publish_master_image(image_name))
+        published = publish_master_image(image_name)
+        image_names = image_names + published
     } else {
         image_names.add(publish_image(image_name + ":$BRANCH_NAME"))
     }
@@ -68,9 +69,11 @@ def build_and_publish() {
  *
  * @param image_names list of images to scan
  */
-def scan_image(image_name) {
-    writeFile file: 'anchore_images', text: image_name
-    anchore name: 'anchore_images', engineRetries: '700'
+def scan_images(image_names) {
+    image_names.each { name ->
+        writeFile file: 'anchore_images', text: name
+        anchore name: 'anchore_images', engineRetries: '700'
+    }
 }
 
 
@@ -81,7 +84,7 @@ def scan_image(image_name) {
  */
 def clean_local_images(image_names) {
     withDockerServer([uri: PCIC_DOCKER]){
-        for(name in image_names) {
+        image_names.each { name ->
             sh "docker rmi ${name}"
         }
     }
@@ -89,6 +92,10 @@ def clean_local_images(image_names) {
 
 
 node {
+    stage('Clean Workspace') {
+        cleanWs()
+    }
+
     stage('Code Collection') {
         checkout scm
         sh 'git fetch'
@@ -107,17 +114,14 @@ node {
     def image_names
     stage('Build and Publish Image') {
         image_names = build_and_publish()
+        echo "$image_names"
     }
 
     stage('Security Scan') {
-        scan_image(image_names[0])
+        scan_images(image_names)
     }
 
     stage('Clean Local') {
         clean_local_images(image_names)
-    }
-
-    stage('Clean Workspace') {
-        cleanWs()
     }
 }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,6 +1,77 @@
...
+
+
+/**
+ * If the master branch has been tagged we also add the `latest` tag.  Otherwise
+ * we just use the branch name as the tag.
+ *
+ * @return tags a list of the tags for the image
+ */
+def get_tags() {
+    String tag = sh (script: 'git tag --contains', returnStdout: true).trim()
```
jameshiebert: While, *in practice* we usually only tag a commit once with a single version, git technically allows an arbitrary number of tags.

```
james@basalt:/tmp/foo$ git log
commit afdc173e87cdfc8cbf4d4c184f2d4da5ac6bcd33 (HEAD -> master, tag: some-other-name, tag: 1.0.0-dev, tag: 1.0.0)
Author: James Hiebert <james@hiebert.name>
Date:   Thu Dec 5 09:14:15 2019 -0800

    First commit
james@basalt:/tmp/foo$ git tag --contains
1.0.0
1.0.0-dev
some-other-name
```

Since you're returning a list already, I think you just need to process the output of `git tag` properly to allow for multiple tags. Then you should be good to go.

Jenkinsfile
```diff
@@ -1,6 +1,77 @@
...
+
+    def tags = []
+    if(BRANCH_NAME == 'master' && !tag.isEmpty()) {
```
jameshiebert: That seems like sensible logic to me.

Jenkinsfile
```diff
@@ -1,6 +1,77 @@
...
+
+    withDockerServer([uri: PCIC_DOCKER]){
+        docker.withRegistry('', 'PCIC_DOCKERHUB_CREDS') {
+            tags.each { tag ->
+                image.push(tag)
+            }
+        }
+    }
+
+    return tags[0]
```
jameshiebert: Why do you only return the first?

jameshiebert:
I'd agree that it's probably a good idea to wait until a branch is in PR to run the security scan. 9 times out of 10, we're not going to need it, and don't want it slowing down development.

> But PRs typically get a lot of commit activity too. Can the security scan be made optional in some way?

Do they? I'd argue that maybe they do in practice, but maybe they *shouldn't*? A good PR should have one set of change pushed once for the initial PR and then one more set pushed to address comments. And even if they *do* see a lot of commit activity, I'd argue that PRs are exactly the time when we want to slow down, be thorough and do all the steps (even if they are inconvenient).

rod-glover:
Quite possibly I misuse PRs; I often use them to do my own "internal" code reviews based on changes from baseline, before I ask for an external review. A PR is also where I often put a TODO list, often arising from my internal reviews. The former can be replaced by the [compare view](https://help.github.com/en/github/committing-changes-to-your-project/comparing-commits-across-time), although I can't make comments to myself in it. The latter I'd have either to put in the issue backing the PR or do without.

I definitely not opposed to a change in how we use PRs, just pointing out their convenience as they are currently (ab)used.

Jenkinsfile
nikola-rados: That isn't something I considered, good catch!

Jenkinsfile
nikola-rados: So in the security scan I just want to scan the image that was just pushed. To identify it I just need one of the tags, so arbitrarily I choose the first one. I updated the function to return the full list and then just use the first item later on.

nikola-rados:
So the changes I have pushed address the following:
- handle multiple git tags (`get_tags`)
- only conduct security scan on branches filed as pull requests (`Security Scan Stage`)
- use consistent branch naming (`get_branch_name`)

For the branch naming, I got feedback from @corviday that once a branch is filed as a PR the tag changes from the branch name to `PR-[pull request #]`.  To handle this I just added a quick check for the original name before tagging the image.

If there are any other changes I should make let me know!

jameshiebert:
LGTM!

PR #355, opened 2020-02-10
Use getScanName feature

nikola-rados:
The pipeline was not specifying a tag in the `imageName` so Anchore was just using the `latest` tag as it's scan target.  This was not the desired behaviour as we would want to scan the image that was just created.  This PR introduces the shared library pipeline step `getScanName` which returns the appropriate image name and tag to scan.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,9 +12,10 @@ node {
 
     def image
     def imageName
+    def imageSuffix = 'climate-explorer-frontend'
 
     stage('Build Image') {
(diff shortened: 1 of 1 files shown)


nikola-rados:
Scans will now be conducted on `master` as well as `PR-#` branches.

nikola-rados:
This PR is no longer valid due to a new [feature in `jenkins-library`](https://github.com/pacificclimate/jenkins-library/pull/11)

PR #359, opened 2020-02-12
Fix DataspecSelector construction of options

### LINKED ISSUE #358: Map dataset selection does not offer any choices
Currently, the dataset selection dropdown in the map settings dialogue does not offer any choices other than the currently selected one.

map_datasets

To reproduce, select any data parameters and open the map settings menu.

rod-glover:
Resolves #358


rod-glover:
And likewise on dev instance.

PR #364, opened 2020-03-20
Add new docker build args

nikola-rados:
The `Dockerfile` uses an environment variable that is created by a script in the repo.  The changes here run that script, capture the output and pass it into the build command.


rod-glover:
This looks great, @nikola-rados ! Nice and simple and unless I am out to lunch, should do the job.

PR #367, opened 2020-05-14
Add workflow to github actions

nikola-rados:
This PR adds a workflow that is used by github actions.  

The workflow is responsible for:
- running the node test suite in multiple versions (in parallel)
- building, tagging and publishing a docker image
- conducting an anchore scan on PR's and master

I think it would be a worthwhile experiment to drop travis and jenkins in favour of actions in this repo.  The capabilities of actions has really grown over the past couple of months and continues to grow.  Some of the features used in this workflow were not available just a few months ago.  Furthermore, I feel this repo does not have an overly complex CI pipeline which makes it a good candidate to test out a new tool.


nikola-rados:
In order to allow for some more flexibility I have broken down the single CI file into three separate workflows.  Each has slightly different conditions for triggering and I think they are more readable this way

- CI tests will run on every push
- Image building, tagging and publish will happen on each push + if a tag has been made
- Image scan will only happen on PRs targeted to merge to master and once the docker image has been published

The build is currently failing on the `anchore scan`.  Fixing those vulnerabilities is out of scope for this PR so this is good to merge.

PR #376, opened 2020-07-07
Update workflows

nikola-rados:
The github action workflows have been updated to bring them within out actions standards.

- Trigger adjustments
- Workflow name updates
- Updated scanning/tagging strategy
- Updated `README.md`
- Remove `Jenkins` & `Travis`


.github/workflows/image-scan.yml
```diff
@@ -1,26 +1,26 @@
...
 
-on:
-  pull_request:
-    branches:
-      - master
+on: push
+  # push:
+  #   branches:
+  #     - master
+  #     - '*security*'
```
corviday, line 7: Why is this commented out? Is this a step we used to do but aren't doing any more?

README.md
```diff
@@ -145,9 +146,9 @@ If you *really* want to skip the linting during a commit, you can always run `gi
 
 ### Notes
 
-#### Jenkins
+#### Github Actions
 
-Jenkins automatically handles the generation of docker images.  Currently it is configured to trigger an image build for each push on individual branches.  The image generated will have the name `climate-explorer-frontend/[branch_name]`.
+The workflows setup in actions will automatically build, tag and publish to our [docker hub](https://hub.docker.com/r/pcic/climate-explorer-frontend).
 
 ##### Configuration, environment variables, and Docker
```
corviday, line 153: I think this should be a four-hash headline too; it's not "under"  `Github Actions`, right?

.github/workflows/image-scan.yml
nikola-rados: That shouldn't look like that. Thanks for pointing it out.

PR #383, opened 2020-08-17
Fix data / axis mismatch in Variable Response Graph

### LINKED ISSUE #329: Axes on variable response graph are sometimes reversed
reversedaxes

This is not correct. PRSN values should be around 1. fdETCCDI values should be around 20. Labels on the wrong axes?

corviday:
This fixes an issue in the construction of the variable response graph. The graph "pairs up" data from multiple variable timeseries, to make a time-independent comparison of the variables. For example, for the variables `a` and `b` and timeseries `[t0, t1, t2]`:
 
The input timeseries would be:
```
[(t0, a0), (t1,  a1), (t2, a2)] 
[(t0, b0), (t1, b1), (t2, b2)]
```

And the output would be:
```
[(a0, b0), (a1, b1), (a2, b2)]
```

However, there was a bug where the output of the "pairing up" step was actually:
```
[(b0, a0), (b1, a1), (b2, a2)]
```

Which is valid variable response data, but all the axis labels and general graph formatting expected `a` to be the x-axis and `b` to be the y-axis. The result was graphs with data axes mislabeled. Here's a graph from the live website:
wrongaxes

As you can see, this graph show mean precipitation of 25 mm per day, and mean tasmax values of less than 2 degrees Celsius. The data are on the wrong axes. Here is the same graph, output by this branch:

fixedaxes

Resolve #329

### DIFF
--- a/src/core/chart-transformers.js
+++ b/src/core/chart-transformers.js
@@ -72,20 +72,20 @@ function makeVariableResponseGraph (x, y, graph) {
   
   const xseries = _.filter(graph.data.columns, series => seriesNameContains(series, x));
   const yseries = _.filter(graph.data.columns, series => seriesNameContains(series, y));
-    
+
   let tuples = [];
   let seriesMatched = false;
-  for(let dependent of xseries) {
+  for(let independent of xseries) {
     //Try to match each dependent variable series with an independent variable series
-    let independent = _.find(yseries, series => {
+    let dependent = _.find(yseries, series => {
       return series[0].toLowerCase().replace(y.toLowerCase(), x.toLowerCase()) === 
-        dependent[0].toLowerCase();
+        independent[0].toLowerCase();
       });
-    if(independent) {
+    if(dependent) {
       seriesMatched = true;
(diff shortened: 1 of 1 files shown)


corviday:
Huh, I assumed you used `let` for loop variables because they changed values between loops. I'll swap them out.

rod-glover:
That's what I meant by counterintuitive.

corviday:
I'm getting errors. 

```
  ● makeTimeSliceGraph › generates a yearly timeslice

    TypeError: Assignment to constant variable.

      373 |   }
      374 | 
    > 375 |   for(const i = 0; i < graph.data.columns.length; i++) {
          |                                                    ^
      376 |     let series = graph.data.columns[i];
      377 |     if(!_.isUndefined(series[sliceIndex]) && series[0] !== 'x'){
      378 |       slicedData.push([series[0], series[sliceIndex]]);
```

So you should use `const` when the `for` loop iterates over a collection with `in`, but not for `for` loops that iterate numerically with `++`, is that correct?

PR #386, opened 2020-10-02
Update Help to reflect expanded DD and RP availability

corviday:
Updates the help to match the new Return Period and Degree Day datasets. Return period now has all six standard climatologies available (previously it had only four, and the projected ones were a year off) and both data collections are now available for all Canada.

### DIFF
--- a/public/external-text/default.yaml
+++ b/public/external-text/default.yaml
@@ -594,7 +594,7 @@ help:
         2010-2039,
         2040-2069, and
         2070-2099.
-        This data is available for all of British Columbia.
+        This data is available for all of Canada.
 
         A degree-day is a measure of how much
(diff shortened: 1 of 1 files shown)


public/external-text/default.yaml
```diff
@@ -646,13 +646,15 @@ help:
...
         1971-2000,
-        2011-2040,
-        2041-2070, and
-        2071-2100.
-        This data is available for all of British Columbia.
+        1981-2010,
+        2010-2039,
+        2040-2069, and
+        2070-2099.
+        This data is available for all Canada.
```
rod-glover, line 657: With six items, I'd consider making them a list rather than part of a sentence.

rod-glover (approved):
LGTM, with a tiny question.

PR #388, opened 2020-10-28
Fix climate layers

### LINKED ISSUE #387: Map no longer showing climate layers
As of version 2.1.0, the map no longer displays the selected climate layer. Regressing to 2.0.5 fixes it.

Investigation reveals that we are no longer passing correct information to component `DataLayer`; specifically its parameter `layerParams` is incorrect, and more specifically `layerParams` lacks the `dataset` property. It may lack others too.

rod-glover:
Resolves #387 

Very simple fix, converting a string to a number. 

It would be interesting to know if the CE backend metadata format changed recently, since if memory serves this code was working a couple of months ago. CE backend has seen some changes since then, and this might have been one of them, possibly unintended.

### DIFF
--- a/src/components/map-controllers/map-helpers.js
+++ b/src/components/map-controllers/map-helpers.js
@@ -37,7 +37,8 @@ export function hasComparand(props) {
 
 // predicate that detects whether a timestamp index is a 0th index:
 // either January, winter, or the first year in a file. 
-export const is0thIndex = timestamp => (JSON.parse(timestamp).timeidx === 0);
+export const is0thIndex = timestamp =>
+  Number(JSON.parse(timestamp).timeidx) === 0;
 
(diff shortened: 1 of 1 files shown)


rod-glover:
[Demo](http://docker-dev01.pcic.uvic.ca:30509/)

corviday (approved):
Well, that is certainly a pleasantly straightforward solution. I too wonder when / how this became a bug.

jameshiebert:
> I too wonder when / how this became a bug.

Probably [here](https://github.com/pacificclimate/climate-explorer-frontend/commit/94e187bbe652b1350e849c05d199a7a276adf7af).

rod-glover:
Well spotted, James. I look out for those implicit type conversion gotchas, but missed that one. Thanks.

PR #394, opened 2020-11-27
Reduce docker footprint

### LINKED ISSUE #393: Reduce actions footprint
- Remove multi os actions
- Upgrade to 20.04
- Remove PR trigger

nikola-rados:
Resolves #393.

The anchore image scan action has been updated to reflect the suggested layout for `v2`. To note, `v2` is experiencing a large amount of false positives at the moment. The nice thing though is that `v2` saves the reports as issues in the security tab which is great for tracking.


rod-glover (approved):
This looks quite plausible, with the caveat that I lack deep knowledge.

Issue #397, opened 2021-02-04
Missing Long Term Average timestamp on initial load


corviday:
To reproduce:

Go to PCEX, click on the tab for the Long Term Average Graph. Graph will not show up, because the frontend is making a request with no time index, specifically https://services.pacificclimate.org/pcex/api/data?ensemble_name=ce_files&model=PCIC12&variable=cdd&emission=historical,rcp85&area= - no time parameter at all.

This doesn't happen if you select a dataset from the dropdown, or a timestamp from the graph dropdown. It's only missing on initial load.

corviday:
This bug seems to be somewhat intermittent, occurring maybe one time in ten? So it is probably the result of a race condition. It may not, as I said in the initial comment, depend on whether a selection is made prior to rendering the graph.

Normally, graph components upon loading initially send a "bad" request with no time parameter - this is documented in a separate issue [here](https://github.com/pacificclimate/climate-explorer-frontend/issues/384) - followed immediately by a "good" request with a time parameter set.

This implies that the mechanism of the bug _isn't_ that the time parameter is unset, but that the "good" request is never sent after the  "bad" request. Perhaps something about the handling of the "bad" request disrupts the second request. In that case, one possible solution to explore is trying to detect the "bad" request and not send it.

PR #398, opened 2021-03-05
Add a new data portal for flood frequency data

### LINKED ISSUE #391: Add ability to select a single point on the map
Currently, the PCEX map allows you to draw a shape to select the area of interest, from which data is displayed.  We want to display some streamflow data, where the area of interest is typically single point. The backend already supports returning data from a WKT point, so we should just be able to add a point selection tool to the map.

corviday:
Creates a new portal for hydrological flood frequency data.

The new portal has: 
* an app controller, `FloodAppController`, that is extremely similar to `SingleAppController`, differing only in its name and its child components
* a data controller, `FloodDataController`. This controller has only one graph (long term average) and the statistical table, because this data is so simple there's only one way to "slice" it. There's only a single model (ensemble of models, actually) so we don't need the Context Graph or Snapshot graph, and all data is annual-only, so we don't need the annual cycle graph or the annual anomaly graph. Only the Long Term Average graph is present. The next PR will supplement the Long Term Average with additional series that show the ensemble percentiles.
* a map controller. It is using `SingleMapController`, but a new boolean `prop` has been added, `selectPoint`. When `selectPoint` is `true`, users will indicate an area of interest by selecting a single point, as is appropriate for hydrology data. When `selectPoint` is `false`, users will indicate an area of interest by drawing a rectangle or polygon, the current behaviour, which will be preserved in all other portals.

[Demo](http://docker-dev01.pcic.uvic.ca:30400/pcex/app/#/data/flood/upper_fraser)

Resolves #391

### DIFF
--- a/src/components/app-controllers/FloodAppController/FloodAppController.js
+++ b/src/components/app-controllers/FloodAppController/FloodAppController.js
@@ -1,0 +1,230 @@
+/***************************************************************
+ * FloodAppController.js 
+ * 
+ * This controller displays streamflow return period data.
+ * Its differences from regular data display (SingleAppContoller) 
+ * are:
+ *  - all data is annual (different graphs)
+ *  - displays ensemble percentiles
+ *  - point selectionon map
+ ***************************************************************/
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Col, ControlLabel, Grid, Panel, Row } from 'react-bootstrap';
+
+import SingleMapController from '../../map-controllers/SingleMapController';
+import FloodDataController
+  from '../../data-controllers/FloodDataController/FloodDataController';
+import {
+  EmissionsScenarioSelector,
+  ModelSelector,
+  VariableSelector,
+} from 'pcic-react-components';
+import {
+  datasetFilterPanelLabel,
+  emissionScenarioSelectorLabel,
+  modelSelectorLabel,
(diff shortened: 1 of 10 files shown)


corviday:
This is affected by [this bug](https://github.com/pacificclimate/climate-explorer-frontend/issues/397).

src/components/app-controllers/FloodAppController/FloodAppController.js
```diff
@@ -0,0 +1,231 @@
...
+        ensemble_name={ensemble_name(this.props)}
+        {...this.state}
+        onChangeArea={this.handleChangeArea}
+        onChangeModel={this.handleChangeModel}
+        onChangeScenario={this.handleChangeScenario}
+        onChangeVariable={this.handleChangeVariable}
+      />
+    );
+  }
+}
```
rod-glover, line 231: Plausible, boilerplate. Anything that might need a closer look?

rod-glover (approved):
LGTM, app works fine. Didn't dive super deep, but it looks quite straightforward and plausible.

src/components/app-controllers/FloodAppController/FloodAppController.js
corviday: Not really, it's all copy and paste from SingleAppController.

PR #407, opened 2021-04-30
Add watershed information table

### LINKED ISSUE #405: More context on streamflow point selection
It's quite difficult to tell what point you have selected in the Streamflow Extremes portal. The map doesn't actually give you any feedback about what latitude and longitude you are selecting.

Short term, we could make the latitude and longitude more user visible, such as by displaying the latitude and longitude the user has clicked on somewhere next to the graph, and/or including it in the downloadable CSV, and/or displaying it under the map when the user is making a selection like the PDP does. Then users can double-check their selection, or know what points to put into the watershed API to get information on the watershed.

Long term, we might wish to have the front end access the watershed API and display that information directly to the user. We could put the numerical information (area, elevation, melton ratio) in a table below the map, and we could display the JSON outline of the selected watershed directly on the map. I don't think this would be too hard.

corviday:
This PR addresses a UI issue with the new streamflow portal. The streamflow portal has the user click on a single point on the map (streamflow does not make sense averaged across an area) but it can be difficult for the user to tell whether they have indeed clicked on the point they're interested in; the UI doesn't offer any feedback. These changes add a table showing some information about the watershed that drains to the point the user clicked on: the latitude and longitude of the selected point (from the map) and information provided by the `watershed` API: the area of the watershed that drains to that point, highest and lowest elevations, and the Melton ratio. 

Changes:

- Adds `AttributeValueTable`, a table that will display a collection of objects with `attribute`, `value` and `units` members
- Adds `WatershedSummaryTable`, a component that queries the `watershed` backend for data and passes it to an `AtributeValueTable`
- `getWatershed()` and `getWatershedGeographyName()` in `ce-backend.js` for querying the watershed API
- `validateWatershedData()` and `parseWatershedTableData()` in `util.js` for checking the API response and formatting it into attribute/value/unit objects for the table
- some linting in `util.js` and `ce.js`, but the functions explicitly mentioned above are the only really new code
- tests for the new stuff, which I think is more linecount than the new code

Resolves #405


src/components/guidance-content/info/InformationItems.js
```diff
@@ -652,6 +652,26 @@ export const exportStatsTableDataLabel = (
...
+  <LabelWithInfo label='Watershed Information'>
+    <p>
+      This table presents information about the watershed that drains
+      to a point selected on the map.
+    </p>
+    <p>
+      The Outlet Latitude, Longitude, and Elevation describe the point selected on the map.
+      The Source Elevation, Area, and Melton Ratio describe the watershed drained by streams
+      that empty to the selected point. Source Elevation is the highest point in the watershed.
+      Area is the total area of the watershed,
```
corviday, line 668: s/,/./

src/components/AttributeValueTable/AttributeValueTable.js
```diff
@@ -0,0 +1,52 @@
...
+          >
+            Units
+          </TableHeaderColumn>
+        </BootstrapTable>
+      </div>
+    );
+  }
+}
+
+export default AttributeValueTable;
```
rod-glover, line 52: Totally minor: You can fold this into the class defn above: `export default class ...`.

src/components/WatershedSummaryTable/WatershedSummaryTable.js
```diff
@@ -0,0 +1,157 @@
...
+import {
+  parseWatershedTableData,
+  validateWatershedData,
+} from '../../core/util';
+import { errorMessage } from '../graphs/graph-helpers';
+
+import styles from './WatershedSummaryTable.module.css';
+
+
+// TODO: Use `withAsyncData` to factor out common data-fetching code here
```
rod-glover, line 28: Do you feel like trying this? Only if you have time.

src/core/__test_data__/sample-API-results.js
```diff
@@ -26,7 +26,7 @@ import _ from 'lodash';
 
 //A set of monthly, seasonal, and annual to test time resolutions:
 
-var monthlyTasmaxTimeseries = {
+const monthlyTasmaxTimeseries = {
```
rod-glover, line 29: :+1:

src/core/__tests__/util-test.js
```diff
@@ -53,6 +53,52 @@ jest.mock('../../data-services/public.js');
...
+        
+    },
+  ];
+  
+  describe('parseWatershedTableData', function () {
+     it('parses a watershed API response into attribute-value-units tuples', function () {
+         var result = util.parseWatershedTableData(mockAPI.watershed, "POINT (-119.15625 53.09375)");
+         expect(result).toEqual(watershedTableTestExpected);
+     }); 
+  });
```
rod-glover, line 101: :+1:

src/core/__tests__/util-test.js
```diff
@@ -182,6 +228,27 @@ jest.mock('../../data-services/public.js');
       expect(util.validateUnstructuredTimeseriesData({data: concatenatedTasmaxTimeseries})).toEqual({data: concatenatedTasmaxTimeseries});
     });
   });
+  
+  describe('validateWatershedData', function () {
+    it('rejects Workzeug error messages', function () {
+      var func = function () {util.validateUnstructuredTimeseriesData({data:
```
rod-glover, line 234: `const`, please? Also could do `const func = () => { ... }`; briefer code, seems to be preferred in a lot of JS work lately.

src/core/__tests__/util-test.js
```diff
@@ -182,6 +228,27 @@ jest.mock('../../data-services/public.js');
...
+        _.forEach(attributes, function(att) {
+            const missing = _.omit(mockAPI.watershed, att);
+            const func = function () {util.validateWatershedData({data: missing})}
+            expect(func).toThrow();
+        });
+    });
+    it('accepts a valid watershed', function () {
+        expect(util.validateWatershedData({data:mockAPI.watershed})).toEqual({data: mockAPI.watershed});
+    })
+  });
```
rod-glover, line 251: Wow, awesome testage.

src/core/util.js
```diff
@@ -450,3 +494,22 @@ export function nestedAttributeIsDefined(o, ...attributes) {
   }
   return true;
 }
+
+/**********************************************************
+ * Geo-related helper function
+ **********************************************************/
+// Converts a single WKT point string to a geoJSON point object.
+// There are libraries that do this sort of thing, but right now,
+// this is the only such conversion we need.
```
rod-glover, line 503: :+1:

rod-glover (approved):
Nice job. Great testing. (Isn't it amazing how more than half the effort of development is in creating tests?)

A couple of thoughts about the UI:
- Is it possible to change the hint popup on the circle marker in the map toolbar to read something like "Select a point on the map"? I recall looking into this for some other map tool, but can't remember if it proved intransigent. Alternatively, change the prompt in the unselected state to read "Select a point on the map, using the circle marker tool, to see watershed information." 
- The watershed information is great. May I suggest making it more explicit that the watershed in question is the one draining to the selected point? So maybe a change to the heading of the table, and/or revising the label of the Area and Melton Ratio params to be explicit about this?

corviday:
These are both great UI suggestions, thanks!

>     * Is it possible to change the hint popup on the circle marker in the map toolbar to read something like "Select a point on the map"? I recall looking into this for some other map tool, but can't remember if it proved intransigent. Alternatively, change the prompt in the unselected state to read "Select a point on the map, using the circle marker tool, to see watershed information."

It definitely seems like it *should* be possible to set a custom tooltip - "Draw a circlemarker" isn't very intuitive - but my efforts have come to naught. I updated the table text instead, as you suggested.

>     * The watershed information is great. May I suggest making it more explicit that the watershed in question is the one draining to the selected point? So maybe a change to the heading of the table, and/or revising the label of the Area and Melton Ratio params to be explicit about this?

Done, thanks!

PR #413, opened 2021-06-21
Revert JSZip version to 2.x

### LINKED ISSUE #361: Export to shapefile is not working
Currently, a user can download a shape in any format except shapefile, which produces the following console error messages:

```
Uncaught TypeError: Cannot read property '0' of undefined
    at poly.js:88
    at Array.reduce (<anonymous>)
    at Object.a [as parts] (poly.js:85)
    at t.exports (write.js:21)
    at geo.js:34
    at Array.forEach (<anonymous>)
    at geo.js:32
    at j (geo.js:66)
    at Object.save (geo.js:151)
    at t.value (GeoExporterDialog.js:21)
```

To reproduce:
1. Draw a shape on the map
2. Click on the "Export Polygon" button at the side of the map.
3. Select "Shapefile"

Possibly related: #250

corviday:
JSZip was [refactored](https://stuk.github.io/jszip/documentation/upgrade_guide.html) between version 2.x, which we were previously using, and version 3.x. Some of the functions we were using, like `zip.generate()` are not present in 3.x, which made it impossible to download a shape drawn on the map as a shapefile.

This PR pins JSZip to 2.5 and fixes a too-narrow check for empty coordinates that would treat coordinates consisting of a list of empty lists as if they were were defined coordinates to be written to the shapefile, leading to an error.

[Demo](http://docker-dev01.pcic.uvic.ca:30066/pcex/app/#/data/climo/ce_files)

Resolves #361

### DIFF
--- a/src/core/geo.js
+++ b/src/core/geo.js
@@ -30,7 +30,7 @@ function createZippedShapefile(gj, options) {
 
   [geojson.point(gj), geojson.line(gj), geojson.polygon(gj)]
   .forEach(function(l) {
-    if (l.geometries.length) {
+    if (_.flattenDeep(l.geometries).length > 0) {
       write(
         // field definitions
         l.properties,
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "enzyme": "^3.6.0",
     "enzyme-adapter-react-16": "^1.5.0",
     "filesaver.js": "^0.2.0",
-    "jszip": "^3.2.1",
+    "jszip": "2.5.0",
     "leaflet": "^1.3.4",
     "leaflet-draw": "^1.0.4",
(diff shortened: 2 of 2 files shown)


src/core/geo.js
```diff
@@ -30,7 +30,7 @@ function createZippedShapefile(gj, options) {
 
   [geojson.point(gj), geojson.line(gj), geojson.polygon(gj)]
   .forEach(function(l) {
-    if (l.geometries.length) {
+    if(_.flattenDeep(l.geometries).length > 0){
```
rod-glover, line 33: Good catch!

Tiny point: Our style prefers spaces before and after the parentheses in if statements.

rod-glover (approved):
LGTM

PR #425, opened 2022-10-28
Display selected watershed on map

corviday:
This PR does the following:

* when a user selects the outlet of a watershed on the map, make a call to the watershed API and display the outline of the watershed on the map
* switch from the Upper Fraser dataset to the entire Fraser dataset (different data ensembles in the base URL)

Demo here: https://services.pacificclimate.org/dev/pcex/app/#/data/flood/fraser

### DIFF
--- a/src/components/DataMap/DataMap.js
+++ b/src/components/DataMap/DataMap.js
@@ -397,6 +397,8 @@ class DataMap extends React.Component {
                 showLength: false,
               },
             }}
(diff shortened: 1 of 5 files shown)


rod-glover (approved):
LGTM. I would suggest black for the outline of upstream polgyon; it's hard to see the blue against many of the raster's blue-ish colours.

Issue #447, opened 2023-02-07
UI updates for Extreme Streamflow portal map


corviday:
The Extreme Streamflow portal's map behaves differently than the other maps - instead of drawing and area and seeing the average of some climactic variable across that area, you select a outlet point and can see streamflow at that point. A couple UI changes to make the functionality of this map clearer:

- [ ] Remove the "import polygon" tool entirely - this portal doesn't allow polygons!
- [ ] Change the "draw a circle marker" tooltip to say "Select an outlet"

corviday:
Changing the name of a map tool seems to be impossible in the library we're using. See [this bug report](https://github.com/alex3165/react-leaflet-draw/issues/66) for that library.

corviday:
Things that don't initialize the circle marker tool with the desired tooltip:

```
circlemarker: "Select an outlet",
```

```
circlemarker: allowGeometryDraw && this.props.pointSelect && 
  "Select an outlet",
```

```
circlemarker: allowGeometryDraw && this.props.pointSelect && { title: 'Select an outlet point'},
```

```
 circlemarker: allowGeometryDraw && this.props.pointSelect && {text: 'Select an outlet point'},
```

PR #450, opened 2023-03-24
Updates to extreme streamflow portal help

### LINKED ISSUE #448: Update Help for Extreme Streamflow Portal
Hydrological Impacts will supply some new verbiage for the PCEX help - add it to the .yaml.

corviday:
Updates to help text for the Extreme Streamflow Portal. Resolves #448 , makes progress on #447

Have not yet completed requested updates to the Filtered Datasets Summary, Data Map, and Time of Year Selector due to these components being shared across multiple portals with different desired help texts. The way forward on this is probably to pass a string to each component telling it which portal it is being used by so it can load appropriate help. (Though completely removing the Time Of Year selector would be even better - this portal uses only annual data).

### DIFF
--- a/public/external-text/default.yaml
+++ b/public/external-text/default.yaml
@@ -702,19 +702,20 @@ help:
           
         ### Annual maximum streamflow quantiles
 
-        The streamflow quantiles describe extreme streamflow events
+        The streamflow quantiles describe daily extreme streamflow events
         that would be expected to occur once during a specified 
         "return period," for example once every 20 years. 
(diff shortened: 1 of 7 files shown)


corviday:
Realized I could use the `pointSelect` prop, which determines whether or not the user is permitted to select a point (as in the streamflow portal) to determine which help to display for the map. So that's working now.

Remaining un-updated help is the Filtered Datasets Summary, and the Time of Year Selector.

PR #465, opened 2024-06-03
Implement UI changes/fixes requested by RCI

### LINKED ISSUE #343: Change from Baseline graph does not respond to Dataset selector
The Dataset selector on CfB may be supposed to set the baseline for the %change graphs. It doesn't.

Alternatively, the baseline is fixed, and this selector is superfluous. 

Figure out which is true; then make the required changes.

@corviday as we discussed you have some information to add to this.

eyvorchuk:
This PR contains the UI changes/fixes RCI requested while experimenting with the MBCn climatologies page. These are the changes:

- Replaces instances of `Yearly` with `Annual`
- Represents the `PCIC12` dataset with a thick black line in the `Model Context` graph regardless of the selected model
- Shows the graph tooltip for the selected model in the `Model Context` graph.
- Instead of using exact dates in the x-axis labels of the `Long Term Average` and `Model Context` graphs, they use decadal periods (e.g. `2020s` instead of `2025-07-02`). This is because for some periods (like 1981-2010), some models have `1996-07-02` as their middle date while others have `1997-07-02`, causing the dates to overlap in the `Model Context` graph. This change ensures that the label is the same for all models in each period.
- Removes `Snapshot` graph since the information is already in the `Model Context` graph
- In the `Filtered Datasets Summary` and `Data Map`, only includes the `run_id` if the selected model has multiple runs (e.g. MIROC5 has `r1i1p1` and `r3i1p1`)
- Enables the historical baseline in the `Change from Baseline` graph to be selectable by the dropdown, using 1981-2010 as the default (also resolves #343)

Demo available at https://services.pacificclimate.org/dev/pcex/app/#/data/climo/ce_files

### DIFF
--- a/src/components/DataSpecSelector/DataSpecSelector.js
+++ b/src/components/DataSpecSelector/DataSpecSelector.js
@@ -3,6 +3,8 @@ import PropTypes from "prop-types";
 import Selector from "../Selector/Selector";
 import _ from "lodash";
 import { datasetSelectorLabel } from "../guidance-content/info/InformationItems";
+import { hasMultiRuns } from "../map-controllers/map-helpers";
+
 /******************************************************************
  * DataSpecSelector.js - Data Specification selecting widget
  *
@@ -49,7 +51,7 @@ export default class DataSpecSelector extends React.Component {
       _.sortBy(ids, (item) => item[1]),
       (item) => item[1],
     );
-    if (ids[0][1].split(" ")[0] == ids[ids.length - 1][1].split(" ")[0]) {
+    if (!hasMultiRuns(this.props.meta)) {
       // Only one run id for this model. Remove from labels.
       ids.forEach((item) => {
         item[1] = item[1].split(" ")[1];
--- a/src/components/DataSpecSelector/DataSpecSelector.js
+++ b/src/components/DataSpecSelector/DataSpecSelector.js
@@ -49,6 +49,16 @@ export default class DataSpecSelector extends React.Component {
       _.sortBy(ids, (item) => item[1]),
       (item) => item[1],
     );
+    ids = _.sortedUniqBy(
+      _.sortBy(ids, (item) => item[1]),
+      (item) => item[1],
+    );
+    if (ids[0][1].split(" ")[0] == ids[ids.length - 1][1].split(" ")[0]) {
+      // Only one run id for this model. Remove from labels.
+      ids.forEach((item) => {
+        item[1] = item[1].split(" ")[1];
+      });
+    }
     return ids;
   }
 
--- a/src/components/DataSpecSelector/DataSpecSelector.js
+++ b/src/components/DataSpecSelector/DataSpecSelector.js
@@ -49,10 +49,6 @@ export default class DataSpecSelector extends React.Component {
       _.sortBy(ids, (item) => item[1]),
       (item) => item[1],
     );
-    ids = _.sortedUniqBy(
-      _.sortBy(ids, (item) => item[1]),
-      (item) => item[1],
-    );
     if (ids[0][1].split(" ")[0] == ids[ids.length - 1][1].split(" ")[0]) {
       // Only one run id for this model. Remove from labels.
       ids.forEach((item) => {
--- a/src/components/data-presentation/FilteredDatasetsSummary/FilteredDatasetsSummary.js
+++ b/src/components/data-presentation/FilteredDatasetsSummary/FilteredDatasetsSummary.js
@@ -25,6 +25,13 @@ export default class FilteredDatasetsSummary extends React.Component {
     dual: false,
   };
 
+  removeSingleRun = (keyedData, meta) => {
+    // Remove ensemble member from keyedData if there is only one run in the metadata
+    if (keyedData && meta.length && !hasMultiRuns(meta)) {
+      keyedData.forEach((el) => (el.key = el.key.split(" ")[1]));
+    }
+  };
+
   render() {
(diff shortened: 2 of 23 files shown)


src/components/DataSpecSelector/DataSpecSelector.js
```diff
@@ -49,6 +49,16 @@ export default class DataSpecSelector extends React.Component {
       _.sortBy(ids, (item) => item[1]),
       (item) => item[1],
     );
+    ids = _.sortedUniqBy(
+      _.sortBy(ids, (item) => item[1]),
+      (item) => item[1],
+    );
```
rod-glover, line 55: It looks like `ids = _.sortedUniqBy(...` is duplicated. Is that intended?

src/components/DataSpecSelector/DataSpecSelector.js
```diff
@@ -49,6 +49,16 @@ export default class DataSpecSelector extends React.Component {
...
+    if (ids[0][1].split(" ")[0] == ids[ids.length - 1][1].split(" ")[0]) {
```
rod-glover, line 56: Never use `==` comparison; its type-coercion is too tricky and leads to errors. Use only strict equality `===`. This is standard advice in JS land.

rod-glover: Also, it's hard to deduce what is being compared here and why. Suggest you decompose the `ids` items into helpfully-named values that you then compare. Something like

```
const first_run_id = ids[0][1].split(" ")[0];
const last_run_id = ids[ids.length - 1][1].split(" ")[0];
if (first_run_id === last_run_id) {
   ...
```

If you were feeling really fun you could define a temp function `const run_id = id = id[1].split(" ")[0]` to DRY things up.

rod-glover: Alternatively, could the condition be evaluated by 

```
import { hasMultiRuns } from "../../map-controllers/map-helpers";
```

as it is in another component further down?

src/components/data-presentation/FilteredDatasetsSummary/FilteredDatasetsSummary.js
```diff
@@ -45,8 +46,20 @@ export default class FilteredDatasetsSummary extends React.Component {
...
+
     const keyedComparandData =
       this.props.comparandMeta && this.props.comparandMeta.map(metaToKeyedData);
+    if (
+      keyedComparandData &&
+      this.props.comparandMeta.length &&
+      !hasMultiRuns(this.props.comparandMeta)
+    ) {
+      keyedComparandData.forEach((el) => (el.key = el.key.split(" ")[1]));
+    }
```
rod-glover, line 62: Much duplication here. DRY up using a function.

rod-glover (changes requested):
Looking quite good. Some comments inline, and the following two questions:

1. 
> Enables the historical baseline in the Change from Baseline graph to be selectable by the dropdown

Can you provide an example case where this dropdown appears? I don't see it in any of the cases I pulled up.

2.
> In the Filtered Datasets Summary and Data Map, only includes the run_id if the selected model has multiple runs (e.g. MIROC5 has r1i1p1 and r3i1p1)

I don't see any difference in the label or the content for Filtered Datasets Summary. `run_id` is included/excluded with MIROC5/PCIC12 respectively in Data Map, but not in Filtered Datasets Summary.

eyvorchuk:
@rod-glover 
> Can you provide an example case where this dropdown appears? I don't see it in any of the cases I pulled up.

Could we meet to see what you're looking at? I just opened the [demo](https://services.pacificclimate.org/dev/pcex/app/#/data/climo/ce_files) and I was able to see the dropdown in the `Change From Baseline` graph.
> I don't see any difference in the label or the content for Filtered Datasets Summary. `run_id` is included/excluded with MIROC5/PCIC12 respectively in Data Map, but not in Filtered Datasets Summary.

Sorry, I should've worded this better for the `Filtered Datasets Summary`. Only the `Label in Selectors` column excludes the `run_id` in cases where a model has only one run. Currently in production, for PCIC12 from 1971-2000 as an example, that column displays the label as `rXi1p1 1971-2000`, but with this change, the label will be `1971-2000` instead.

rod-glover:
> @rod-glover
> 
> > Can you provide an example case where this dropdown appears? I don't see it in any of the cases I pulled up.
> 
> Could we meet to see what you're looking at? I just opened the [demo](https://services.pacificclimate.org/dev/pcex/app/#/data/climo/ce_files) and I was able to see the dropdown in the `Change From Baseline` graph.

I see now that the "Dataset" dropdown is what we're talking about. I misinterpreted it and thought it chose which future dataset (or something) would be displayed. For an inexpert (or inattentive) user, I think a better label here would be "Baseline".

> 
> > I don't see any difference in the label or the content for Filtered Datasets Summary. `run_id` is included/excluded with MIROC5/PCIC12 respectively in Data Map, but not in Filtered Datasets Summary.
> 
> Sorry, I should've worded this better for the `Filtered Datasets Summary`. Only the `Label in Selectors` column excludes the `run_id` in cases where a model has only one run. Currently in production, for PCIC12 from 1971-2000 as an example, that column displays the label as `rXi1p1 1971-2000`, but with this change, the label will be `1971-2000` instead.

Ah, understood now.

rod-glover (approved):
I see you have changed the label in code from "Dataset" to "Baseline", although it doesn't seem to have been deployed yet. I'll trust it all works! Nice work.

PR #468, opened 2024-12-10
Extend Fraser/Peace/Columbia streamflow and fix the crash loop

Nospamas:
- Reworked loading variables to be inline with [react query's reccomendations](https://tanstack.com/query/v4/docs/framework/react/guides/queries), will make it easier if we integrate it later.
---
Secondary changes:
- Does some package updates (audit fix, prettier)
- Adds some VSCode launch capabilities for debugging
---
Available for testing on [beehive](https://beehive.pacificclimate.org/pcex/app/#/data/flood/fraser_peace_columbia)


jameshiebert:
I read over your code changes and it all looks good. The state changes are the bulk of the work... they're a bit verbose, but if they align with the react-state recommendations, then I'm happy :)

I'm always suspicious of any PR that does multiple things, and this one does the thing that I initially set out to do (extend the streamflow frontend to another watershed) and the bitrot problem (the crash loop).

Would you be able to rebase your commits on to our main branch and confirm that the crash loop is fixed under those circumstances? Then add the watershed extension commits?

src/components/WatershedSummaryTable/WatershedSummaryTable.js
```diff
@@ -120,7 +129,7 @@ export default class WatershedSummaryTable extends React.Component {
     }
 
     // Waiting for data
-    if (this.state.fetchingData || this.state.data === null) {
+    if (!this.state.data) {
```
jameshiebert, line 132: This appears to work great both at the outset and if a user clears a previously selected watershed outlet.

Nospamas:
> I read over your code changes and it all looks good. The state changes are the bulk of the work... they're a bit verbose, but if they align with the react-state recommendations, then I'm happy :)
> 
> I'm always suspicious of any PR that does multiple things, and this one does the thing that I initially set out to do (extend the streamflow frontend to another watershed) and the bitrot problem (the crash loop).
> 
> Would you be able to rebase your commits on to our main branch and confirm that the crash loop is fixed under those circumstances? Then add the watershed extension commits?

I've cherry picked out those commits unrelated to the fraser river extension into #469

PR #471, opened 2024-12-17
Deprecate use of process.env for runtime config

Nospamas:
- Replace process.env with window.env
- Move env variables into a config.js file
- Rework docker build to use file
- Rework docker build to use static built files
- Rework docker compose to load a config file
- Add makefile utility matching station-data-portal
- Add healthcheck to docker file
- Updated documentation

Matches configuration that was done as part of [station-data-portal#186](https://github.com/pacificclimate/station-data-portal/pull/186)


Nospamas:
deployed at https://beehive.pacificclimate.org/pcex-fast/app/ for testing.

.github/workflows/docker-publish.yml
```diff
@@ -13,15 +13,20 @@ jobs:
...
+        run: |
+          npm ci
+          npm run build
       - name: Publish to Registry
         uses: docker/build-push-action@v1
         with:
           username: ${{ secrets.pcicdevops_at_dockerhub_username }}
           password: ${{ secrets.pcicdevops_at_dockerhub_password }}
+          dockerfile: docker/Dockerfile
           repository: pcic/climate-explorer-frontend
```
jameshiebert, line 30: Are you concerned about the deprecation warning on this?
https://github.com/pacificclimate/climate-explorer-frontend/pull/471/checks#step:6:1

And maybe that would address [this warning](https://github.com/pacificclimate/climate-explorer-frontend/pull/471/checks#step:6:20).

.github/workflows/docker-publish.yml
```diff
@@ -13,15 +13,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Export REACT_APP_CE_CURRENT_VERSION env var
+      - name: Export REACT_APP_APP_VERSION env var
         run: |
           git fetch --prune --unshallow
-          echo "REACT_APP_CE_CURRENT_VERSION=$(git describe --tags --abbrev=0) ($(git rev-parse --abbrev-ref HEAD):$(git log -1 --format=%h))" >> $GITHUB_ENV
+          echo "REACT_APP_APP_VERSION=$(git describe --tags --abbrev=0) ($(git rev-parse --abbrev-ref HEAD):$(git log -1 --format=%h))" >> $GITHUB_ENV
+      - name: Build npm package
```
jameshiebert, line 20: 2 minutes for the build is *so* much better, especially given that it happens before deployment. Nice!

jameshiebert (approved):
This looks great to me, @Nospamas . I can't say that I completely follow everything, but the performance is better, it cuts a lot of cruft and keeps our dependencies up to date. If you're happy with it, I am.

.github/workflows/docker-publish.yml
Nospamas: Aye, I am concerned by it in so far as I do think we should fix it, but as it predates this issue so I haven't looked into fixing it just yet. I'll roll out these changes to each app (to fix startup times) then I'll go back and review updating that workflow to the latest version.

